PR #6, opened 2019-08-08
re-disable operator

sjenning:
Due to some gaps we uncovered with how to rollout Tech Preview operator with the CVO, we are disabling this in 4.2.  Looking to enable again in 4.3 after coordinating with CVO on design for Tech Preview operators.

@derekwaynecarr @smarterclayton @joelsmith @tkatarki @blomquisg


sjenning:
also @mrunalp

PR #7, opened 2020-01-23
Create default VPA if needed, start all 3 VPA controllers, add recommendation-only mode

joelsmith:
I recommend reviewing this PR on a per-commit basis since each commit provides a different feature

### DIFF
--- a/pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
+++ b/pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
@@ -29,9 +29,7 @@ import (
 )
 
 const (
-	controllerName                  = "vertical-pod-autoscaler-controller"
-	vpaRecommenderServiceAccount    = "vpa-recommender"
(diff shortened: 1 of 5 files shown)


pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
```diff
@@ -326,18 +424,39 @@ func (r *Reconciler) AutoscalerDeployment(vpa *autoscalingv1.VerticalPodAutoscal
 	return deployment
 }
 
-// RecommenderPodSpec returns the expected podSpec for the deployment belonging
+// DefaultVPAController returns a default VerticalPodAutoscalerController instance
+func (r *Reconciler) DefaultVPAController() *autoscalingv1.VerticalPodAutoscalerController {
+	var smf float64 = 0.15
```
rphillips: slack discussion... perhaps move these constants to the top and document why they were selected.

PR #9, opened 2020-02-12
Auto-obtain TLS Key/Cert for Mutating Webhook Service (+ cleanup)

joelsmith:
This PR contains several updates in separate commits:

1. Update RBAC to match latest operand
2. Have OLM install VPA and VPA checkpoint CRDs
   * Also, make VPA controller CRD namespace instead of cluster scoped
   * Also, fix OLM manfiest data, add default channel
3. Use a unique "app" label for each of the three controllers
4. Auto-obtain TLS Key/Cert for Mutating Webhook Service

   To avoid having the manual step of an admin having to generate a TLS
   key/cert pair for use by the Admission Controller's Mutating Webhook
   Service, we needed to re-arrange a lot things:

   * Have the operator create the service, complete with
    annotations which the OpenShift service-ca controller
    sees so that it will generate the key/cert pair and make
    them available in a secret
   * Have the operator create an empty configmap, annotated
    so that the OpenShift service-ca controller will provide
    the CA cert used to create the service's key/cert pair
   * Update the admission controller pod so that it mounts
    the secret with the TLS key/cert pair and also the
    configmap with the CA cert. 
   * Use admission controller args to use cert/key from
      the secret and configmap

   Additonally:
   * fix a small constant capitalization mistake

### DIFF
--- a/pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
+++ b/pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
@@ -30,6 +30,8 @@ import (
 
 const (
 	controllerName = "vertical-pod-autoscaler-controller"
+
+	AdmissionControllerAppName = "vpa-admission-controller"
 	// Fraction of usage added as the safety margin to the recommended request. This default
 	// matches the upstream default
 	DefaultSafetyMarginFraction = 0.15
@@ -44,6 +46,7 @@ const (
 type ControllerParams struct {
 	Command           string
 	NameMethod        func(r *Reconciler, vpa *autoscalingv1.VerticalPodAutoscalerController) types.NamespacedName
+	AppName           string
 	ServiceAccount    string
 	PriorityClassName string
 	GetArgs           func(vpa *autoscalingv1.VerticalPodAutoscalerController, cfg *Config) []string
@@ -55,6 +58,7 @@ var controllerParams = [...]ControllerParams{
 		"recommender",
 		(*Reconciler).RecommenderName,
 		"vpa-recommender",
+		"vpa-recommender",
 		"system-cluster-critical",
 		RecommenderArgs,
 		(*Reconciler).RecommenderEnabled,
@@ -63,6 +67,7 @@ var controllerParams = [...]ControllerParams{
 		"updater",
 		(*Reconciler).UpdaterName,
 		"vpa-updater",
+		"vpa-updater",
 		"system-cluster-critical",
 		UpdaterArgs,
 		(*Reconciler).UpdaterEnabled,
@@ -70,6 +75,7 @@ var controllerParams = [...]ControllerParams{
 	{
 		"admission-controller",
 		(*Reconciler).AdmissionPluginName,
+		AdmissionControllerAppName,
 		"vpa-admission-controller",
 		"system-cluster-critical",
 		AdmissionPluginArgs,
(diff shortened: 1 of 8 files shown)


joelsmith:
@rphillips I'll sync with you tomorrow, but could you PTAL?
/assign @rphillips

pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
```diff
@@ -500,6 +702,100 @@ func (r *Reconciler) VPAPodSpec(vpa *autoscalingv1.VerticalPodAutoscalerControll
...
+	spec.Containers[0].Env = append(spec.Containers[0].Env, corev1.EnvVar{
+		Name: "NAMESPACE",
+		ValueFrom: &corev1.EnvVarSource{
+			FieldRef: &corev1.ObjectFieldSelector{
+				FieldPath: "metadata.namespace",
+			},
+		},
+	})
+	spec.Containers[0].Ports = append(spec.Containers[0].Ports, corev1.ContainerPort{
+		ContainerPort: 8000,
```
rphillips: Is this port configurable?

pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
```diff
@@ -500,6 +702,100 @@ func (r *Reconciler) VPAPodSpec(vpa *autoscalingv1.VerticalPodAutoscalerControll
...
+	spec.Containers[0].VolumeMounts = append(spec.Containers[0].VolumeMounts, corev1.VolumeMount{
+		Name:      "tls-ca-certs",
+		MountPath: "/data/tls-ca-certs",
+		ReadOnly:  true,
+	})
+	spec.Volumes = append(spec.Volumes, corev1.Volume{
+		Name: "tls-certs",
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: "vpa-tls-certs",
```
rphillips: Should be `WebhookCertSecretName`

pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
```diff
@@ -500,6 +702,100 @@ func (r *Reconciler) VPAPodSpec(vpa *autoscalingv1.VerticalPodAutoscalerControll
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      WebhookServiceName,
+			Namespace: r.config.Namespace,
+		},
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{
+				{
+					Port:       443,
+					TargetPort: intstr.FromInt(8000),
```
rphillips: is this port configurable?

pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
joelsmith: Good catch. Fix on its way.

pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
joelsmith: It is configurable in the admission controller, but it doesn't need to be configurable by the end user. I think it's okay for us to be opinionated about it, since the only thing connecting to it will be the API server and it will use the service that we create. So the port number doesn't really matter. It might make sense for us to make it a constant though so that it stays in sync. What do you think?

rphillips: sounds good

PR #10, opened 2020-03-12
Update golang.org/x/crypto to remove all doubt about CVE-2020-9283

joelsmith:
I doubt that there's any exploit path, but we'll update so
as not to show up on vulnerability scans.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1808182


joelsmith:
/retest
@rphillips PTAL

PR #11, opened 2020-03-31
Fix inconsistency between RBAC and CSV required permissions

jwforres:
... and also fix a typo in CRB name

I wasn't sure whether the required permissions in the CSV were correct, or the RBAC in the RBAC file were correct, but the RBAC was not providing sufficient permissions and OLM was rejecting installation of the operator.  If the required permissions are what is not correct I can fix those instead.

### DIFF
--- a/install/02_vpa-rbac.yaml
+++ b/install/02_vpa-rbac.yaml
@@ -22,6 +22,7 @@ rules:
   resources:
   - pods
   - nodes
+  - limitranges
   verbs:
   - get
   - list
@@ -177,6 +178,7 @@ rules:
   - batch
   resources:
   - jobs
+  - cronjobs
   verbs:
   - get
   - list
@@ -204,7 +206,7 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: system:vpa-evictionter-binding
+  name: system:vpa-evictioner-binding
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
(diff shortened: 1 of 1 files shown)


joelsmith:
Our original plan was to install via CSV, but now that we will be installing via OLM, the CSV install files are obsolete. I'm planning to remove them as soon as the OLM install works. As I understand it, installation via OLM is blocked because we don't have images built for it yet. If you would like to try out my development images, I can provide instructions on how to install them using a custom catalog source that I built for testing.

jwforres:
@joelsmith I created my own custom CatalogSource off of the manifest content in the /manifests directory in master, that operator content won't deploy as is, it fails requirements on all of the RBAC checks, hence why i tried using the RBAC yaml to set up the RBAC. Do you have a different set of manifests in a development branch?

joelsmith:
I do have a different set of manifests, but the only difference is the images that the CSV references. Instead of `quay.io/openshift/vertical-pod-autoscaler-operator:4.4` it uses `quay.io/joelsmith/openshift-vertical-pod-autoscaler-operator:latest` and instead of `quay.io/openshift/vertical-pod-autoscaler:4.4` it has `quay.io/joelsmith/openshift-vertical-pod-autoscaler:latest`.  I'm not sure that the ones at `quay.io/openshift` are being updated and I wasn't going to worry about it until the ART team gets our build updated to do images for OLM.



The ART team is in the process of switching our images over and until we get the new ones, I haven't been using the ones that they build.  Here's my catalog source image which references my development images of the VPA operator and the VPA:
`quay.io/joelsmith/vpa-registry:latest`
It uses the same RBAC as what is checked in to master.

joelsmith:
/lgtm
/approve
Thanks for the updates!

PR #14, opened 2020-04-24
Bug 1819479: Only create "SuccessfulUpdate" events when updates are made

joelsmith:
Also:
- get release version from deployement object, not VPA controller object
- set pod spec defaults so that DeepEquals returns true for equivalent
  pod spec templates

### DIFF
--- a/pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
+++ b/pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
@@ -257,17 +257,17 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 			klog.Info(msg)
 			continue
 		}
-		if err := r.UpdateAutoscaler(vpa, params); err != nil {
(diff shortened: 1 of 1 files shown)


joelsmith:
@rphillips PTAL when you have a sec. Thanks!
/assign @rphillips

pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
```diff
@@ -257,17 +257,17 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 			klog.Info(msg)
 			continue
 		}
-		if err := r.UpdateAutoscaler(vpa, params); err != nil {
+		if err, updated := r.UpdateAutoscaler(vpa, params); err != nil {
```
rphillips, line 260: let's reverse the order here of the returned error, usually golang does the error last, (updated, err).

joelsmith: Thanks. Updated.

PR #19, opened 2020-05-26
Bug 1803231: Have OLM auto-create ns for VPA

joelsmith:


joelsmith:
@rphillips could you please review?

PR #20, opened 2020-05-28
bug 1803231: image-references: align names with RCM standards

sosiouxme:
I'm pretty sure we're gonna need this; I keep finding out ways this doesn't work how I expect.

I'm testing out builds to make sure. If so, we can probably re-use bug 1803231 since it's the same problem.


joelsmith:
Thanks!
/approve
/lgtm

sosiouxme:
yeah... it seems to be working.

PR #25, opened 2020-06-03
Add required CSV fields, update VPA CR example, add VPA Controller CR example

joelsmith:
The last image build failed the metadation validation step. See http://external-ci-coldstorage.datahub.redhat.com/cvp/cvp-redhat-operator-metadata-validation-test/ose-vertical-pod-autoscaler-operator-metadata-container-v4.5.0.202006031723.dev-1/ca074d41-4ded-4a5f-912a-1bfaaf6188c7/operator-metadata-linting-output.txt

This PR adds the required and recommended fields.

To validate the manifests, I did:
```
$ pip3 install operator-courier
$ operator-courier verify --ui_validate_io manifests/
```
which is what the metadata import process does as part of its tests.

PR #26, opened 2020-06-03
[release-4.5] bug 1803231: Add required CSV fields, update VPA CR example, add VPA Controller CR example

joelsmith:
This is a cherry-pick of #25


joelsmith:
master branch fix has merged, but we don't have a way to test a 4.6 bug since image builds aren't set up for 4.6 and we can't afford to wait for them. Manually adjusting bz labels.

PR #32, opened 2020-07-21
manifests: change 4.5 to 4.6

sosiouxme:
I was putting this off until I could improve how this is done, but got busy.

Our appregistry/bundle automation is looking for everything in your manifests to say 4.6... I think this is the change needed.


sosiouxme:
our builds seem to fine now, thanks!

joelsmith:
Thanks for updating it

PR #38, opened 2020-10-07
bug 1882524: [release-4.6] manifests/art.yaml: CSV name is verticalpodautoscaler

sosiouxme:
Having successfully pushed to stage once, tests began failing, revealing that the CSV name replacement that was supposed to happen to give it a unique name for each build - wasn't matching. So it gets the same name every time which fails the test. This should make the match work. This is required to ship.


joelsmith:
Thanks!
/approve
/lgtm

PR #48, opened 2020-12-03
bug 1904260: CSV: make version match what art.yaml looks for

sosiouxme:
This version needs to change each time it's released;
without this it will stay `1.0.0` forever and fail tests.

We will need the same thing for 4.7+ (though it needs to be `4.7.0` there).


joelsmith:
/lgtm
/approve
Thanks!

eparis:
manually overriding valid-bug because ART still isn't building 4.7 artifacts for QA to test.

PR #53, opened 2021-03-11
Bring manual install RBAC in line with CSV RBAC

joelsmith:


joelsmith:
I think that the e2e-aws test is flaky right now.
/retest

PR #54, opened 2021-03-12
Bug 1899423: Fix VPA Updater

joelsmith:
Updater now requires the `NAMESPACE` env var to be set so that it can make sure
the admission controller is running. Otherwise, it will emit this message in
its logs:

```
updater.go:116] Error getting Admission Controller status: leases.coordination.k8s.io "vpa-admission-controller" not found. Skipping eviction loop
```

This PR makes it so all VPA controllers set the `NAMESPACE` environment variable, where previously only the admission plugin set it.


joelsmith:
@wangchen615 I think this should fix the issue that you ran into. Once this merges to master branch, we will need to cherry-pick it to the 4.7 branch since 4.7 has the same problem. It's too bad that I didn't pay more attention to [Bug 1899423](https://bugzilla.redhat.com/show_bug.cgi?id=1899423) which was opened in November.

PR #56, opened 2021-03-12
Bump 4.7 to 4.8

joelsmith:
Now that 4.7 has been branched, master should use version 4.8.


jupierce:
File name needs to change as well `vertical-pod-autoscaler.v4.7.0.clusterserviceversion.yaml` -> `vertical-pod-autoscaler.v4.8.0.clusterserviceversion.yaml`

PR #57, opened 2021-03-13
Sync OLM CRDs to manual-install CRDs

joelsmith:
Also split `install/06_vpacheckpoint-crd.yaml` out of `install/05_vpa-crd.yaml`


joelsmith:
@wangchen615 sure enough, the manual-install CRDs were out of date. I'll be opening another PR which adds a check to make sure that they stay updated in the future so that we don't waste our time using outdated versions of the CRDs or RBAC.

PR #60, opened 2021-03-17
add Makefile target e2e-olm-local e2e-local e2e-olm-ci e2e-ci to automate the olm installation and the upstream e2e testing

### LINKED ISSUE #59: Automate the olm installation and e2e testing
The operator should have a makefile target that automates the build, the operator's OLM installation, and the upstream e2e testing. 

The current e2e testing script should point to the upstream repo and have a `full-vpa` e2e testing for the default VPA operator.

/assign @wangchen615

wangchen615:
1. Added dev-image, dev-push targets in Makefile to allow dev image build.
2. Add deploy-olm-local, deploy-olm-ci, deploy to deploy the operator in local mode, in ci mode, and in non-olm mode. Correspondingly, artifacts and images are added for various deployment options.
3. Updated hack/e2e.sh to add the upstream e2e testing for the operator.
4. Automate the e2e testing in e2e-olm-local and e2e-local targets.

Resolve and close #59


Makefile
```diff
@@ -5,25 +5,72 @@ REPO_PATH   ?= $(ORG_PATH)/$(PROJECT)
...
+HASDOCKER := $(shell command -v docker 2> /dev/null)
 ifeq ($(NO_DOCKER), 1)
   DOCKER_CMD =
   IMAGE_BUILD_CMD = imagebuilder
 else
-  DOCKER_CMD := podman run --rm -v "$(CURDIR):/go/src/$(REPO_PATH):Z" -w "/go/src/$(REPO_PATH)" openshift/origin-release:golang-1.15
-  IMAGE_BUILD_CMD = buildah bud
+	ifndef HASDOCKER
+	  DOCKER_CMD := podman run --rm -v "$(CURDIR):/go/src/$(REPO_PATH):Z" -w "/go/src/$(REPO_PATH)" openshift/origin-release:golang-1.12
+	  IMAGE_BUILD_CMD = buildah bud
```
rphillips, line 61: since the podman and docker commands are the same, we can set RUNTIME (or something) to podman or docker names.

Makefile
```diff
@@ -5,25 +5,72 @@ REPO_PATH   ?= $(ORG_PATH)/$(PROJECT)
...
+	  IMAGE_PUSH_CMD = podman push
+	else
+	  DOCKER_CMD := docker run --rm -v "$(CURDIR):/go/src/$(REPO_PATH):Z" -w "/go/src/$(REPO_PATH)" openshift/origin-release:golang-1.12
+	  IMAGE_BUILD_CMD = docker build
+	  IMAGE_PUSH_CMD = docker push
+	endif
 endif
 
+# build image for ci
+CI_REPO ?=registry.svc.ci.openshift.org
```
rphillips, line 71: probably should be defined at the top of the file with the other variables.

Makefile
```diff
@@ -5,25 +5,72 @@ REPO_PATH   ?= $(ORG_PATH)/$(PROJECT)
...
 NO_DOCKER ?= 0
+HASDOCKER := $(shell command -v docker 2> /dev/null)
 ifeq ($(NO_DOCKER), 1)
   DOCKER_CMD =
   IMAGE_BUILD_CMD = imagebuilder
 else
-  DOCKER_CMD := podman run --rm -v "$(CURDIR):/go/src/$(REPO_PATH):Z" -w "/go/src/$(REPO_PATH)" openshift/origin-release:golang-1.15
-  IMAGE_BUILD_CMD = buildah bud
+	ifndef HASDOCKER
+	  DOCKER_CMD := podman run --rm -v "$(CURDIR):/go/src/$(REPO_PATH):Z" -w "/go/src/$(REPO_PATH)" openshift/origin-release:golang-1.12
```
rphillips, line 60: make sure we are using golang 1.15... we are reverting to 1.12 here.

test/registry-setup/main.go
```diff
@@ -0,0 +1,283 @@
...
+func withCI(object *corev1.ConfigMap) error {
+	if err := populate(getCIImageURL, object); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func validateWithOLM(object *corev1.ConfigMap) error {
+	if object.Data[oldOperatorImageEnv] == "" {
```
rphillips, line 164: these map lookups should probably check for the existance of the key.

test/registry-setup/main.go
```diff
@@ -0,0 +1,283 @@
...
+	return
+}
+
+func read(path string) (object *corev1.ConfigMap, err error) {
+	reader, openErr := os.Open(path)
+	if err != nil {
+		err = fmt.Errorf("unable to load file %s: %s", path, openErr)
+		return
+	}
+
```
rphillips, line 259: need a `defer reader.Close()`

test/registry-setup/main.go
```diff
@@ -0,0 +1,283 @@
...
+	image, err := getter()
+	if err != nil {
+		return err
+	}
+
+	oldOperatorImageURL := os.Getenv(oldOperatorImageEnv)
+	oldOperandImageURL := os.Getenv(oldOperandImageEnv)
+	csvFilePath := os.Getenv(csvFilePathEnv)
+
+	if len(object.Data) == 0 {
```
rphillips, line 203: this implies the object already exists, I don't believe we need the object.Data = map[string]string{} on the next line

Makefile
```diff
@@ -5,25 +5,72 @@ REPO_PATH   ?= $(ORG_PATH)/$(PROJECT)
...
+	else
+	  DOCKER_CMD := docker run --rm -v "$(CURDIR):/go/src/$(REPO_PATH):Z" -w "/go/src/$(REPO_PATH)" openshift/origin-release:golang-1.12
+	  IMAGE_BUILD_CMD = docker build
+	  IMAGE_PUSH_CMD = docker push
+	endif
 endif
 
+# build image for ci
+CI_REPO ?=registry.svc.ci.openshift.org
+$(call build-image,vertical-pod-autoscaler-operator,$(CI_IMAGE_REGISTRY)/autoscaling/vertical-pod-autoscaler-operator,./images/ci/Dockerfile,.)
```
joelsmith, line 72: does this run the build-image target unconditionally? I'd prefer to associate this only with targets that need it

Makefile
```diff
@@ -97,3 +152,104 @@ vet: ## Apply go vet to all go files
 .PHONY: help
 help:
 	@grep -E '^[a-zA-Z/0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+clean:
+	rm -rf $(OUTPUT_DIR)
+	oc delete mutatingwebhookconfigurations vpa-webhook-config || true
+	oc delete ns openshift-vertical-pod-autoscaler || true
+	oc delete crd verticalpodautoscalercheckpoints.autoscaling.k8s.io verticalpodautoscalercontrollers.autoscaling.openshift.io verticalpodautoscalers.autoscaling.k8s.io || true
```
joelsmith, line 160: I would rather these oc commands be part of a separate target. Typically `clean:` is just for removing build artifacts and I think people would be surprised to see it trying to talk to their cluster.

Maybe they could be in a target called `clean-deploy:` or something like that?

Also, maybe it would be good to use a `$(KUBECTL)` variable like you do for other targets, instead of `oc`

hack/e2e.sh
```diff
@@ -4,9 +4,37 @@ set -euo pipefail
 
 GOPATH="$(mktemp -d)"
 export GOPATH
+AUTOSCALER_PKG="k8s.io/autoscaler"
```
joelsmith, line 7: I would prefer it if we could have this test aim at a specific fork of the `openshift/kubernetes-autoscaler` repo. That way we can have OpenShift-specific patches if we need, and by using a specific fork, older branches of the CI will target older branches of the VPA. We want to avoid having broken CI on an old release branch when in the future, it tries to use the latest VPA, which might have a test for a new feature which the old release branch doesn't support.

If you would like, I can work on some makefile code to handle this case.

joelsmith (changes requested):
This all looks very good!

If possible, I would like to see existing `install/*` or `manifests/*` files used rather than adding new copies of those to `artifacts/`. If the new makefile targets need modified versions of the existing files, I would prefer to see it make a temporary local copy as part of its build (perhaps in the output directory). For files we don't already have like the operator group or subscription, I think it would be fine to add those to the repo, but I don't like the name `artifacts` for that purpose. To me `artifacts` gives the impression of something created by the build.  Maybe `assets`? I'm not sure what the best name is.

wangchen615 (commented):
> This all looks very good!
> 
> If possible, I would like to see existing `install/*` or `manifests/*` files used rather than adding new copies of those to `artifacts/`. If the new makefile targets need modified versions of the existing files, I would prefer to see it make a temporary local copy as part of its build (perhaps in the output directory). For files we don't already have like the operator group or subscription, I think it would be fine to add those to the repo, but I don't like the name `artifacts` for that purpose. To me `artifacts` gives the impression of something created by the build. Maybe `assets`? I'm not sure what the best name is.

I used `install/*`

rphillips:
Let's squash the PR and then I'll lgtm

joelsmith:
@wangchen615  I think we're ready to merge this, but would you please rebase to the current master and squash the commits first?

wangchen615:
> @wangchen615 I think we're ready to merge this, but would you please rebase to the current master and squash the commits first?
> 
> edit: @rphillips beat me to it! 😄

@rphillips @joelsmith the commits are squashed. Thanks.

joelsmith (approved):
Thanks for your hard work to get the e2e tests working!

rphillips:
Thank you again. :)

PR #64, opened 2021-03-19
Allow control of VPA updater paramter minReplicas

joelsmith:
For https://issues.redhat.com/browse/OCPNODE-487

### DIFF
--- a/pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
+++ b/pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
@@ -47,6 +47,8 @@ const (
 	DefaultPodMinMemoryMb = 250
 	// By default, the VPA will not run in recommendation-only mode. The Updater and Admission plugin will run
 	DefaultRecommendationOnly = false
+	// By default, the updater will not kill pods if they are the only replica
+	DefaultMinReplicas = 2
 )
 
 type ControllerParams struct {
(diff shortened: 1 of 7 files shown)


joelsmith:
@wangchen615 please review when you have a moment

pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
```diff
@@ -47,6 +47,8 @@ const (
 	DefaultPodMinMemoryMb = 250
 	// By default, the VPA will not run in recommendation-only mode. The Updater and Admission plugin will run
 	DefaultRecommendationOnly = false
+	// By default, the updater will not kill pods if they are the only replica
+	DefaultMinReplicas = 2
```
wangchen615, line 51: Why put 2 instead of 1 for the `DefaultMinReplicas`? Will by default we have two updaters for just redundancy?

wangchen615 (approved):

pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
joelsmith: The parameter controls how many replicas are required before the updater will kill a pod. It is a cluster-wide setting. So for example, if the value is 2, and I have a VPA operating on my deployment which has just one replica, then the updater will not kill my only pod, even if the recommender thinks its resource requests are very wrong. The idea is that we don't want the VPA to cause downtime for an application.

We're exposing this value now because some customers have requested the ability to set the value to 1. In other words, they want to allow the updater to kill the only pod in a deployment (for example) even though that might mean their application will have downtime.  If they change the value to 1, they will need to set an update policy of `Initial` on any VPAs where they don't want the updater to kill their only pod.

wangchen615: Thanks for the explanation, @joelsmith.

wangchen615 (approved):

PR #65, opened 2021-03-20
Bump dependencies version

QiWang19:
Update go module dependencies and remove Gopkg files.
Add context argument under pkg/operator since introduced by new dependency (github.com/openshift/client-go imports  k8s.io/client-go).

PR #66, opened 2021-03-22
Bump CI tests base image to golang 1.15

joelsmith:


joelsmith:
#65 will already bump these, and it's not the cause of the failing tests in that PR. It looks like we need to bump it in the CI jobs in openshift-release.

PR #69, opened 2021-04-02
Add e2e tests for recommendationOnly field

wangchen615:
1. add await_for_controllers() function in hack/e2e.sh to check if the replica status of 3 controllers change when patch verticalpodautoscalercontroller `default` with `recommendationOnly: true` set. If controllers are not created or deleted appropriately, the script will exit with 1.
2. enable both [sig-autoscaling] [VPA] [full-vpa] e2e test for `recommendationOnly: false` mode and [sig-autoscaling] [VPA] [recommender] [v1] e2e test for `recommendationOnly: true` mode. Any failed upstream e2e tests will exit with 1.

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -2,6 +2,75 @@
 
 set -euo pipefail
 
+function run_upstream_vpa_tests() {
+  if $recommendationOnly
+  then
+    echo "recommendationOnly is enabled. Run the recommender e2e tests in upstream ..."
+    pushd ${SCRIPT_ROOT}/e2e
+    GO111MODULE=on go test -mod vendor ./v1/*go -v --test.timeout=60m --args --ginkgo.v=true --ginkgo.focus="\[VPA\] \[recommender\]" --ginkgo.skip="doesn't drop lower/upper after recommender's restart" --report-dir=/workspace/_artifacts --disable-log-dump
+    V1_RESULT=$?
+    popd
+    echo "v1 recommender test result:" ${V1_RESULT}
+
+    if [ $V1_RESULT -gt 0 ]; then
+      echo "Tests failed"
+      exit 1
+    fi
+  else
+    echo "recommendationOnly is disabled. Run the full-vpa e2e tests in upstream ..."
+    pushd ${SCRIPT_ROOT}/e2e
+    GO111MODULE=on go test -mod vendor ./v1/*go -v --test.timeout=60m --args --ginkgo.v=true --ginkgo.focus="\[VPA\] \[full-vpa\]" --report-dir=/workspace/_artifacts --disable-log-dump
+    V1_RESULT=$?
+    popd
+    echo "v1 full-vpa test result:" ${V1_RESULT}
+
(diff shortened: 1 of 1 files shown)


hack/e2e.sh
```diff
@@ -2,6 +2,81 @@
...
+      exit 1
+    fi
+  fi
+}
+
+function await_for_controllers() {
+  local retries=${1:-10}
+  while [ ${retries} -ge 0 ]; do
+    recommenderReplicas=$(kubectl get deployment vpa-recommender-default -n openshift-vertical-pod-autoscaler -oyaml|yq ".status.replicas")
+    if [[ "$recommenderReplicas" == "null" ]]; then
```
joelsmith, line 38: Also, if you change to the jsonpath version, it will return an empty string instead of `null` if the field doesn't exist, so this entire `if` clause can be replaced with:
```
recommenderReplicas=${recommenderReplicas:=0}
```
Which will cause it to be zero if null or unset.

hack/e2e.sh
```diff
@@ -13,32 +88,19 @@ mkdir -p ${GOPATH}/src/k8s.io
...
-  if [ $V1_RESULT -gt 0 ]; then
-    echo "Tests failed"
-    exit 1
-  fi
-fi
\ No newline at end of file
+  echo "error - not all controllers are running!"
+  exit 1
+fi
+recommendationOnly=$(kubectl get VerticalPodAutoScalerController default -n openshift-vertical-pod-autoscaler -oyaml|yq ".spec.recommendationOnly" || false)
```
joelsmith, line 105: In the case that `kubectl` fails, do you want the script to exit? Or is the `|| false` supposed to put `false` into `recommendationOnly`?

If you want the script to exit, I think you can simply remove `|| false` since it will already exit due to the `-e` shell setting. With or without `|| false` it will exit.

If you wanted a `false` value for `recommendationOnly`, then the end should probably be `|| echo false` instead of `|| false`. That would cause the string `false` to be in `recommendationOnly` if the `kubectl` fails, and the script would continue running.

joelsmith (commented):
If you decide to switch from yaml output + yq to jsonpath output, then there are a few places you could change. If you prefer it as-is, that's fine too. The `|| false` spots where you set `recommendationOnly` might be correct or a bug, depending upon what you intended.

hack/e2e.sh
wangchen615: > In the case that `kubectl` fails, do you want the script to exit? Or is the `|| false` supposed to put `false` into `recommendationOnly`?
> 
> If you want the script to exit, I think you can simply remove `|| false` since it will already exit due to the `-e` shell setting. With or without `|| false` it will exit.
> 
> If you wanted a `false` value for `recommendationOnly`, then the end should probably be `|| echo false` instead of `|| false`. That would cause the string `false` to be in `recommendationOnly` if the `kubectl` fails, and the script would continue running.

I wanted a false value for recommendationOnly, then the end should probably be || echo false. I will change accordingly.

joelsmith:
Looks great! I'll add `lgtm` after squash

PR #72, opened 2021-04-07
Better handling of NO_DOCKER=1 in check scripts

joelsmith:
If `NO_DOCKER=1` is set, check scripts should not try to run in a container.

Also, add `make help` text for `make clean` and `make clean-deploy`


wangchen615 (approved):
/lgtm

PR #75, opened 2021-04-07
Add manifest-diff test to ensure manual-install RBAC & CRD matches OLM

joelsmith:
To run the test:
make manifest-diff


wangchen615 (approved):
/lgtm, thanks. Looks great.

PR #79, opened 2021-04-08
Remove yq dep by using yaml2json + jq instead

joelsmith:


wangchen615 (approved):
/lgtm

Thanks so mucn.

PR #81, opened 2021-04-08
Update Makefile

wangchen615:
skip 04_vpacontroller.yaml creation for e2e-ci as it takes time for the CR to be available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,7 +140,7 @@ test: ## Run unit tests
 
 .PHONY: test-e2e
 test-e2e: ## Run e2e tests
-	hack/e2e.sh $(KUBECTL)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -140,7 +140,7 @@ test: ## Run unit tests
 
 .PHONY: test-e2e
 test-e2e: ## Run e2e tests
-	hack/e2e.sh $(KUBECTL)
+	hack/e2e.sh ${KUBECTL}
```
joelsmith, line 143: I think this change is wrong. For a Makefile, I think we want parens, not braces.

joelsmith: nevermind, either works in POSIX and GNU make

PR #85, opened 2021-04-09
Update Makefile

wangchen615:
Change dev operator and operand image names to be consistent with the CI images used.


rphillips:
hmm. would this change impact upgrades by changing the name?

wangchen615:
> hmm. would this change impact upgrades by changing the name?

No impact on upgrades. This will only change the local build, which will not impact any CI tests. Changing it just to make it look consistent.

PR #89, opened 2021-04-10
Bug 1950210: Update CRD API version to match upstream and cleanup for yamllint

joelsmith:
The CRD API version v1beta1 is being deprecated, so we must update all CRDs to v1.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@
 # ignore the binaries and goland configurations
 .idea
 bin
+
+# output of e2e tests
+/_output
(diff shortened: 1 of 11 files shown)


joelsmith:
This PR has two purposes. In addition to updating the CRD from the deprecated v1beta1 version to v1, I expect this PR to fail the new `manifest-diff` test. Once I see that the test is properly working, I will make the same changes to the `install/deploy` versions of the CRDs.

joelsmith:
@wangchen615 could you please review? I used this to update the upstream object CRDs:

https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml

wangchen615 (approved):
/lgtm

PR #97, opened 2021-06-22
Bug 1974567: Remove unneeded run-level from namespace in manual install manifests

joelsmith:
According to https://bugzilla.redhat.com/show_bug.cgi?id=1805488, only components which must run before the SCC controller starts should use run-level 1. VPA is fine to run with the restricted SCC.

### DIFF
--- a/install/deploy/00_namespace.yaml
+++ b/install/deploy/00_namespace.yaml
@@ -5,4 +5,3 @@ metadata:
   name: openshift-vertical-pod-autoscaler
   labels:
     name: openshift-vertical-pod-autoscaler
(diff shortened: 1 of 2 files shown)


joelsmith:
Chen, could you please take a look?
/assign @wangchen615

wangchen615:
> Chen, could you please take a look?
> /assign @wangchen615

I see the log says the below:
```
Current deployments and pods:
NAME                               READY   UP-TO-DATE   AVAILABLE   AGE
vertical-pod-autoscaler-operator   0/1     0            0           4m38s
No resources found in openshift-vertical-pod-autoscaler namespace.
```

PR #106, opened 2021-09-24
Move VPA from channel per version to stable channel

joelsmith:


wangchen615 (approved):
/lgtm

joelsmith:
/hold cancel
I was too slow to add /hold but we'll revert any/all of these changes if necessary.
The question we still have is about whether on manifests/dirname, does dirname refer to a version or a channel name. So we may need to go back to how we had it before for much of this change. Thanks @wangchen615 for catching that.

PR #114, opened 2021-12-06
Correctness fixes to yaml2json/json2yaml test utils

joelsmith:


joelsmith:
Looks like the cluster didn't fully install on the last run.
/retest

PR #120, opened 2022-04-20
Update VPA CRD to match operand version

joelsmith:


joelsmith:
@wangchen615 I'm surprised to see test failures due to `Asset` and `AssetData` being undefined. I expected your PR https://github.com/openshift/kubernetes-autoscaler/pull/225 would have resolved those test failures. Do you have any idea why it might be failing again?

Also, it looks like bindata has been removed from the vendor dir in the most recent updates to VPA vendor dir. Perhaps we should see about rebasing upstream ASAP to fix that.

wangchen615 (approved):
Thanks.

wangchen615:
> Also, it looks like bindata has been removed from the vendor dir in the most recent updates to VPA vendor dir. Perhaps we should see about rebasing upstream ASAP to fix that.

Yeah, I believe it might be deleted again. We need rebasing upstream ASAP.

PR #129, opened 2022-10-06
Fix stable channel current csv

joelsmith:


joelsmith:
Required by https://github.com/openshift/ocp-build-data/pull/2048

PR #132, opened 2022-10-20
[release-4.10] OCPBUGS-2662: Fix E2E CI tests by making currentCSV match CSV file

joelsmith:
The manifest change only affects a manual build or the builds used for E2E tests. Customer-facing builds use generated manifests and are not affected by this change.

Also, the upstream E2E tests were broken for 4.10 since this fix PR did not merge prior to our upstream rebase for 4.10: https://github.com/kubernetes/autoscaler/pull/4771

So this fix will use the 4.9 code only for the E2E tests (not for anything else).


joelsmith:
This change will need to merge before this PR can pass the tests:
https://github.com/openshift/vertical-pod-autoscaler-operator/pull/113

Also,
/jira refresh

PR #134, opened 2022-10-21
Fix 'go fmt' for switch to golang 1.17

joelsmith:


joelsmith:
Manually adding labels for comments-only changes.

joelsmith:
Required for https://github.com/openshift/vertical-pod-autoscaler-operator/pull/113

PR #135, opened 2022-10-21
Add manifest-diff-upstream to help keep VPA Operator manifests in sync with the VPA operand

joelsmith:
Currently we have no check to detect when a change is made to the upstream VPA which requires a manifest change to the VPA operator. This PR seeks to remedy that deficiency. This new check will fetch 2 key manifests from the OpenShift fork of the Kubernetes Autoscaling repo: The RBAC objects and the CRD objects.

The check detected that we were missing the status stanza of the VPA Checkpoint CRD, so this PR also adds that stanza so that this PR can pass.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,8 @@ vet: ## Apply go vet to all go files
 	hack/go-vet.sh ./...
 
 .PHONY: manifest-diff
-manifest-diff: build-testutil ## Compare permissions and CRDs from OLM manifests and install/*.yaml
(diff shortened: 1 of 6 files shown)


joelsmith:
Here is the sample output of the new test. This is the output of the test before I pushed commit 2 to fix it.
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_vertical-pod-autoscaler-operator/135/pull-ci-openshift-vertical-pod-autoscaler-operator-master-manifest-diff/1583604422619959296/build-log.txt

PR #140, opened 2023-03-03
Update OpenShift VPA docs

joelsmith:
* remove reference to CVO / add reference to OLM
* add link to VPA product docs


joelsmith:
IPI installer failed.
/retest

PR #143, opened 2023-03-24
Add annotation for cluster monitoring

joelsmith:


joelsmith:
There's more work to do here before this can be enabled. The VPA controllers need metrics endpoints and the VPA operator needs to create PodMonitor or ServiceMonitor objects for the metrics endpoints.

PR #150, opened 2023-11-29
PODAUTO-77: PODAUTO-78: Allow container spec overrides for operand containers 

jkyros:
This: 
- Extends the `VerticalPodAutoscalerController` types + CRD to add a `ContinerOverrides` struct that can be used to override container `Resources` and `Args`, which will let people tune their deployments without having to fight/disable the operator
- Adjusts the operator such that it understands the new CRD fields and considers them when generating pod/container specs
- This project was being maintained with a very old (v0.10.0?) operator-sdk which no longer seems to work in a go 1.20 environment, so this grudgingly adjusts the Makefile to generate deepcopy/menifsts the "new" ( > v1.0.0) operator-sdk way until we can do a proper migration to a newer version of operator-sdk


pkg/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
```diff
@@ -757,10 +758,56 @@ func (r *Reconciler) VPAPodSpec(vpa *autoscalingv1.VerticalPodAutoscalerControll
...
+	spec := r.VPAPodSpec(vpa, params)
+
+	// Allow the user to override the resources of the container
+	if (!reflect.DeepEqual(vpa.Spec.ContainerOverrides.Recommender.Resources, corev1.ResourceRequirements{})) {
+		spec.Containers[0].Resources = vpa.Spec.ContainerOverrides.Recommender.Resources
+	}
+
+	// Append user args to our container args
+	if len(vpa.Spec.ContainerOverrides.Recommender.Args) > 0 {
+		spec.Containers[0].Args = append(spec.Containers[0].Args, vpa.Spec.ContainerOverrides.Recommender.Args...)
```
jkyros, line 773: I was appending the user args, but I could probably do something like keda does here if we want it to work like that https://github.com/openshift/custom-metrics-autoscaler-operator/blob/ea0c16951aabdeaecba690d718d2304bc779bc5d/controllers/keda/transform/transform.go#L820

pkg/apis/autoscaling/v1/verticalpodautoscalercontroller_types.go
```diff
@@ -17,7 +18,38 @@ type VerticalPodAutoscalerSpec struct {
...
+	// Admission is the container overrides for the VPA's admission container
+	// +optional
+	Admission ContainerOverride `json:"admission,omitempty"`
+	// Recommender is the container overrides for the VPA's recommender container
+	// +optional
+	Recommender ContainerOverride `json:"recommender,omitempty"`
+	// Updater is the container overrides for the VPA's updater container
+	// +optional
+	Updater ContainerOverride `json:"updater,omitempty"`
+}
```
jkyros, line 39: I'm not sure I like what I called these, but now that I have it generating it's easy to change. 

Do we think this ends like it did over in CMA with being able to override the pod spec too? Do we like that interface e.g.  https://github.com/openshift/custom-metrics-autoscaler-operator/blob/ea0c16951aabdeaecba690d718d2304bc779bc5d/apis/keda/v1alpha1/kedacontroller_types.go#L98 ?

jkyros (commented):
What I did is kind of minimalist, but the VPA operator seems kind of minimalist. We don't have everything we have over in [CMA](https://github.com/openshift/custom-metrics-autoscaler-operator/tree/main/controllers/keda/transform) . I can adjust this to be as...robust...as we think we need.

Makefile
```diff
@@ -91,8 +91,9 @@ depend-update:
 # output paths for the generated CRDs.  It also requires that they
 # already exist in order to regenerate the OpenAPI bits, so we do some
 # copying around here.
-.PHONY: generate
-generate: ## Code generation (requires operator-sdk >= v0.5.0)
+# TODO(jkyros): I renamed this because we're generating the "new" way at the end
+.PHONY: generate-old
+generate-old: ## Code generation (requires operator-sdk >= v0.5.0)
```
jrvaldes, line 96: I recommend moving this refactoring to a separate PR

Makefile
```diff
@@ -293,3 +294,30 @@ operator-registry-image:
 
 	$(IMAGE_BUILD_CMD) --build-arg MANIFEST_LOCATION=$(MANIFESTS_DIR) -t $(LOCAL_OPERATOR_REGISTRY_IMAGE) -f images/operator-registry/Dockerfile.registry .
 	$(DOCKER_RUNTIME) push $(LOCAL_OPERATOR_REGISTRY_IMAGE)
+
+# TODO(jkyros): I can't migrate the whole operator to the new operator-sdk, there's a bunch of stuff we have to re-arrange, the project layout is different,
```
jrvaldes, line 298: Consider creating a Jira card to track this effort in the backlog and link the card here in TODO.

hack/boilerplate.go.txt
```diff
@@ -0,0 +1,15 @@
+/*
```
jrvaldes, line 1: Is this required for this change?

pkg/apis/autoscaling/v1/verticalpodautoscalercontroller_types.go
```diff
@@ -17,7 +18,38 @@ type VerticalPodAutoscalerSpec struct {
 	// +kubebuilder:validation:Minimum=0
 	PodMinMemoryMb     *float64 `json:"podMinMemoryMb,omitempty"`
 	RecommendationOnly *bool    `json:"recommendationOnly,omitempty"`
-	MinReplicas        *int64   `json:"minReplicas,omitempty"`
+	// +kubebuilder:validation:Minimum=1
```
jrvaldes, line 21: Consider mentioning in the commit message why this validation is needed.

jrvaldes (commented):
@jkyros  Thanks for working on this. PTAL at my comments.

pkg/apis/autoscaling/v1/verticalpodautoscalercontroller_types.go
jkyros: I'm not actually changing it, the old CRD had it set, but it was decoupled from the types and I wanted it retained in the CRD generation for compatibility, I'm just truing up the generation with the CRD customers have been using.

hack/boilerplate.go.txt
jkyros: it is if I want the generator be as close to a modern generation as possible yes? Like this was basically "make the new generation method work on the old stuff with as few changes as possible".

pkg/apis/autoscaling/v1/verticalpodautoscalercontroller_types.go
```diff
@@ -17,7 +19,53 @@ type VerticalPodAutoscalerSpec struct {
...
+
+	// We'd love to just use the deployment spec to override container fields, but it's really
+	// unwieldy having a user fill out the container array and have to merge them by name, so we include
+	// a specific container override field.
+
+	// Container allows for direct overrides on the "important" container of the deployment, i.e. the one
+	// that is actually running the operand
+	// +optional
+	Container ContainerOverride `json:"container"`
+}
```
jkyros, line 55: Alright, I left some room for the deployment fields. Right now using it looks like this. Does that feel gross? 
```
apiVersion: v1
items:
- apiVersion: autoscaling.openshift.io/v1
  kind: VerticalPodAutoscalerController
  metadata:
    creationTimestamp: "2023-11-29T18:41:15Z"
    generation: 6
    name: default
    namespace: openshift-vertical-pod-autoscaler
    resourceVersion: "73170"
    uid: 2b544974-2667-42d0-bb50-f8b6d41d291a
  spec:
    deploymentOverrides:
      admission:
        container:
          args:
          - --kube-api-qps=30.0
          - --kube-api-burst=40.0
          resources:
            requests:
              cpu: 40m
              memory: 40Mi
      recommender:
        container:
          args:
          - --kube-api-qps=20.0
          - --kube-api-burst=60.0
          resources:
            requests:
              cpu: 60m
              memory: 60Mi
      updater:
        container:
          args:
          - --kube-api-qps=20.0
          - --kube-api-burst=80.0
          resources:
            requests:
              cpu: 80m
              memory: 80Mi
    minReplicas: 2
    podMinCPUMillicores: 25
    podMinMemoryMb: 250
    recommendationOnly: false
    safetyMarginFraction: 0.15
kind: List
metadata:
  resourceVersion: ""
```

PR #162, opened 2024-03-22
PODAUTO-102: Add NodeSelector and Tolerations fields to deploymentoverrides

joelsmith:
Also, clean up TestOverrideResources's flipped expected/actual ordering

### DIFF
--- a/install/deploy/01_vpacontroller.crd.yaml
+++ b/install/deploy/01_vpacontroller.crd.yaml
@@ -109,6 +109,57 @@ spec:
                                 type: object
                             type: object
                         type: object
+                      nodeSelector:
+                        additionalProperties:
+                          type: string
+                        description: Override the NodeSelector of the deployment's
+                          pod. This allows, for example, for the VPA controllers to
+                          be run on non-master nodes
+                        type: object
+                      tolerations:
+                        description: Override the Tolerations of the deployment's
+                          pod. This allows, for example, for the VPA controllers to
+                          be run on non-master nodes with a specific taint
+                        items:
+                          description: The pod this Toleration is attached to tolerates
+                            any taint that matches the triple <key,value,effect> using
+                            the matching operator <operator>.
+                          properties:
+                            effect:
+                              description: Effect indicates the taint effect to match.
+                                Empty means match all taint effects. When specified,
+                                allowed values are NoSchedule, PreferNoSchedule and
+                                NoExecute.
+                              type: string
+                            key:
+                              description: Key is the taint key that the toleration
+                                applies to. Empty means match all taint keys. If the
+                                key is empty, operator must be Exists; this combination
+                                means to match all values and all keys.
+                              type: string
+                            operator:
+                              description: Operator represents a key's relationship
+                                to the value. Valid operators are Exists and Equal.
+                                Defaults to Equal. Exists is equivalent to wildcard
+                                for value, so that a pod can tolerate all taints of
+                                a particular category.
+                              type: string
+                            tolerationSeconds:
+                              description: TolerationSeconds represents the period
+                                of time the toleration (which must be of effect NoExecute,
+                                otherwise this field is ignored) tolerates the taint.
+                                By default, it is not set, which means tolerate the
+                                taint forever (do not evict). Zero and negative values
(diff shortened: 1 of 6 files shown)


install/deploy/01_vpacontroller.crd.yaml
```diff
@@ -109,6 +109,13 @@ spec:
                                 type: object
                             type: object
                         type: object
+                      nodeSelector:
+                        additionalProperties:
+                          type: string
+                        description: Override the NodeSelector of the deployment's
+                          pod. This allows, for example, for the VPA controllers to
+                          be run on on-master nodes
```
mrunalp, line 117: non-master?

joelsmith: good catch, thanks!

manifests/stable/vertical-pod-autoscaler.clusterserviceversion.yaml
```diff
@@ -547,8 +547,6 @@ spec:
                   runAsNonRoot: true
                   seccompProfile:
                     type: RuntimeDefault
-              nodeSelector:
-                node-role.kubernetes.io/master: ""
```
jkyros, line 551: If we're taking this out of the CSV...in a case where we're not overriding, does the nodeSelector get defaulted to "master" somewhere else in the code/somehow ?

(Like, I assume we want to preserve existing "masters only" behavior for folks who aren't overriding, or do we not?)

pkg/controller/verticalpodautoscaler/verticalpodautoscaler_test.go
```diff
@@ -180,6 +180,48 @@ func TestOverrideArgs(t *testing.T) {
...
+			Operator: corev1.TolerationOpExists,
+		},
+	}
+
+	vpa.Spec.DeploymentOverrides.Admission.Tolerations = tolOverride
+	vpa.Spec.DeploymentOverrides.Recommender.Tolerations = tolOverride
+	vpa.Spec.DeploymentOverrides.Updater.Tolerations = tolOverride
+
+	for _, params := range controllerParams {
+		t.Run(fmt.Sprintf("override %s node selector", params.AppName), func(t *testing.T) {
```
joelsmith, line 218: oops, I copy/pasted the previous test and forgot to rename it. Should be "override %s tolerations"

manifests/stable/vertical-pod-autoscaler.clusterserviceversion.yaml
joelsmith: This is a change in behavior for this feature. It allows the VPA operator to be scheduled anywhere on all clusters. The tolerations are retained so that it can, but is not required to, run on masters. Unfortunately, I'm not aware of a way that we can allow the cluster admin to control where it lands by way of OLM.

joelsmith: Perhaps we should use a node affinity with preferredDuringSchedulingIgnoredDuringExecution so that in clusters where it can run on a master, it will. And we could have a high-weight rule which looks for a vpa-specific node label so that they can label the node where they want the VPA to land, if they really care to control that.

jkyros: Ok cool, thanks. I think that's alright, it's just the operator, if it dies the operand stays alive.

joelsmith: It looks like they have a way to select the operator node via the subscription's spec.config.nodeSelector field. I'll investigate further.

joelsmith: John and I talked it over. We're going to revert the above change, and rely upon an OLM user's ability to override the nodeselector via the subscription config, as explained here:
https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md#nodeselector
That means a cluster admin on a hypershift cluster will need to take the extra step of specifying where the VPA operator should land. We will document this extra step in the OpenShift docs

jkyros:
Thanks Joel, it's awesome that you found a way to keep the old default behavior _and_ give Hypershift what they need. 
/lgtm

PR #172, opened 2024-09-04
PODAUTO-225: Fix small configuration issues

maxcao13:
* Reverts the CSV icon back to a long one line string. (Didn't realize the literal block scalar would break it)
* Because of above ^, ignore the `line-length` rule for csv files
* Fix package name in `Makefile` so `undeploy-bundle` can find the correct operator again.


joelsmith:
Ooh, I like the longer timeout! Thanks for all the clean-up

/lgtm
/approve

PR #173, opened 2024-09-06
PODAUTO-225: Fix e2e-olm-upgrade bugs

maxcao13:
This commit does 3 things:
1. Removes the common app.kubernetes.io/name commonLabel that kustomize uses from the deployment. This is because kustomize commonLabels will mutate the labelSelector that the replica set uses to find its pods. It is discouraged by k8s to update the label selectors so this reverts that change.
2. Adds the `--extra-service-accounts arg` to operator-sdk bundle. This is because during an operator-sdk bundle upgrade, it will delete the service accounts that it thinks do not belong to the CSV anymore (since it did not appear in the CSV).
3. Adds a local-upgrade script that simulates a local upgrade that CI does.

Extra notes on this:

1. https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates notes that label selector changes are discouraged, and kustomize notes that this can be fixed here: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/labels/. Previously, you can just delete the deployment after an upgrade so that the selectors are fixed, but this removes that hassle and confusion as noted here: https://github.com/operator-framework/operator-sdk/issues/5900 and it's linked issues.
2. I completely missed that this [feature was a thing](https://sdk.operatorframework.io/docs/advanced-topics/multi-sa/), and I thought that all the extra files in the bundle directory would suffice to be able to install the operator. But while it does work for an install, the upgrade part would fail because of this issue. This was what was causing the [CI rehearsal to fail here](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/56247/rehearse-56247-pull-ci-openshift-vertical-pod-autoscaler-operator-master-e2e-aws-upgrade/1831876156333232128), because the ServiceAccounts got deleted.


joelsmith:
/lgtm
/approve

Thanks!

PR #176, opened 2024-09-09
[release-4.12] OCPBUGS-40934: update VPA golang.org/x/net for http rapid reset for CVE-2024-8421

joelsmith:
Also update golang to 1.18 for compatibility with golang.org/x/net@v0.23.0. It appears that ART was already building with 1.18, but we were out of sync and still on 1.15 in several places which makes us fail CI tests.

```
sed -i 's/1\.15/1.18/g' Dockerfile Dockerfile.rhel7 Makefile go.mod \
    hack/go-fmt.sh hack/go-lint.sh hack/go-vet.sh hack/manifest-diff.sh \
    images/ci/Dockerfile images/operator-registry/Dockerfile.registry.ci 

go get golang.org/x/net@v0.23.0
go mod tidy
go mod vendor
```


jkyros:
Awesome, yep, ART apparently went all the way to 1.19: 
 https://github.com/openshift/vertical-pod-autoscaler-operator/blob/a4817b7bf9cf63a7634d587cafc5a30dd4940a9f/.ci-operator.yaml#L4 

We were building 1.15 with 1.19, building 1.18 with 1.19 is probably better. :smile: 

/lgtm

PR #177, opened 2024-09-24
PODAUTO-225: Re-enable VPA alm-example in CSV and add spec descriptors

maxcao13:
This commit re-enables the VPA example in the CSV for the OCP console to use. Turns out the VPA is a "supported resource" and operator-sdk adds your supported resource examples to the bundle regardless if it's actually an owned resource or not. This commit simply uncomments the sample, and manually removes it from the bundle folder after we generate the bundle. 

This is what I am talking about: https://github.com/operator-framework/operator-registry/blob/c8307cac2e6ea71b3719e6d5a5850dd639da296f/pkg/lib/bundle/supported_resources.go#L18

Operator-sdk generate bundle goes through files and writes [CRDs, Services, and "supported resources"](https://github.com/operator-framework/operator-sdk/blob/819984d4c1a51c8ff2ef6c23944554148ace0752/internal/cmd/operator-sdk/generate/internal/manifests.go#L40) to the bundle directory only. Unfortunately, we supply one of those resources in an operator-sdk project :p 

This commit also adds spec descriptors to the VPA since the scorecard test now complains about them not being present after adding an example.

Finally, this commit also fixes a bug when deploying the bundle with Make and specifying an OPERAND_IMG.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,10 +243,8 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 
 .PHONY: predeploy
 predeploy: yq ## Setup configuration for the operator before deployment.
-	VPA_OPERAND_PREVIOUS_OP="$(shell $(YQ) eval '.patches[0].patch' config/manager/kustomization.yaml | sed 's/\"/\\\"/g')"
-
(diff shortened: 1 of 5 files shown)


maxcao13:
If anyone wants to test how the spec descriptors actually display, they need to have a cluster and log onto the console and deploy the bundle. Then go to installed operators and try creating a VerticalPodAutoscaler CR.

joelsmith:
Thanks for getting to the bottom of that!
/lgtm
/approve

PR #178, opened 2024-10-18
OCPBUGS-45372: Update default resource requests and QPS/Burst

maxcao13:
Updates defaults according to vpa benchmarking spike/new documentation defaults.

I was thinking of not setting CPU limits for the operator, but since we [_should ideally_](https://sdk.operatorframework.io/docs/best-practices/managing-resources/#managing-resources) set memory limits at least, I added additional CPU limits to the main container (alongisde the kube-rbac-proxy container limits), in case something _really_ goes wrong somehow, we do not use more CPU than intended (and all the operator is doing is reconciling anyways).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -290,6 +290,7 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.2.1
 CONTROLLER_TOOLS_VERSION ?= v0.14.0
+ENVTEST_VERSION ?= release-0.18
 
 # GOFLAGS is explicitly empty here so that we do not run into https://github.com/golang/go/issues/45811 on CI
 # This is set in kustomize, controller-gen, and envtest targets
@@ -311,7 +312,7 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 .PHONY: envtest
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
-	test -s $(LOCALBIN)/setup-envtest || GOFLAGS='' GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.18
+	test -s $(LOCALBIN)/setup-envtest || GOFLAGS='' GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@$(ENVTEST_VERSION)
 
 .PHONY: operator-sdk
 OPERATOR_SDK ?= $(LOCALBIN)/operator-sdk
@@ -451,7 +452,7 @@ undeploy-catalog: ## Undeploy the catalog image.
 	cd $(OLM_OUTPUT_DIR) && $(KUSTOMIZE) edit set image quay.io/openshift/origin-vertical-pod-autoscaler-operator-catalog=$(CATALOG_IMG)
 	$(KUSTOMIZE) build $(OLM_OUTPUT_DIR) | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
 	$(MAKE) delete-ns
-
+	$(MAKE) uninstall
 
 .PHONY: create_vpa_controller_cr
 create_vpa_controller_cr: ## Create a VPA CR.
--- a/bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
+++ b/bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
@@ -55,7 +55,7 @@ metadata:
     categories: OpenShift Optional
     certifiedLevel: Primed
     containerImage: quay.io/openshift/origin-vertical-pod-autoscaler-operator:4.18.0
-    createdAt: "2024-09-25T02:20:09Z"
+    createdAt: "2024-12-03T20:22:23Z"
     description: An operator to run the OpenShift Vertical Pod Autoscaler. Vertical
       Pod Autoscaler (VPA) can be configured to monitor a workload's resource utilization,
       and then adjust its CPU and memory limits by updating the pod (future) or restarting
@@ -688,6 +688,7 @@ spec:
                     fieldRef:
                       fieldPath: metadata.namespace
                 image: quay.io/openshift/origin-vertical-pod-autoscaler-operator:4.18.0
+                imagePullPolicy: Always
                 livenessProbe:
                   httpGet:
                     path: /healthz
@@ -702,6 +703,9 @@ spec:
                   initialDelaySeconds: 5
                   periodSeconds: 10
                 resources:
+                  limits:
+                    cpu: 100m
+                    memory: 500Mi
                   requests:
                     cpu: 20m
                     memory: 50Mi
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -37,6 +37,7 @@ spec:
         - /manager
         args:
         - --leader-elect
+        imagePullPolicy: Always
         env:
         - name: VPA_OPERAND_IMAGE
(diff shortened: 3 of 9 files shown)


bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
```diff
@@ -702,6 +703,9 @@ spec:
                   initialDelaySeconds: 5
                   periodSeconds: 10
                 resources:
+                  limits:
+                    cpu: 100m
+                    memory: 100Mi
```
jkyros, line 708: Are we sure this is high enough that nobody will hit this? Like if they're set to auto-upgrade it and they're over the limit, they'd pull it and die to OOM-kill, right?

internal/controller/verticalpodautoscaler/admission_plugin.go
```diff
@@ -41,5 +53,12 @@ func AdmissionPluginArgs(vpa *v1.VerticalPodAutoscalerController, cfg *Config) [
 		TLSCACertFileArg.Value("/data/tls-ca-certs/service-ca.crt"),
 		WebhookTimeout.Value("10"),
 	}
+	if !CheckArgExists(s.DeploymentOverrides.Admission.Container.Args, KubeAPIQPSArg) {
+		args = append(args, KubeAPIQPSArg.Value("25.0"))
+	}
+	if !CheckArgExists(s.DeploymentOverrides.Admission.Container.Args, KubeAPIBurstArg) {
+		args = append(args, KubeAPIBurstArg.Value("50.0"))
+	}
+
```
jkyros, line 62: 1. Am I correct in interpreting this as "assert default QPS/Burst centrally via the operator rather than letting the operands set defaults" ? 
2. And I was going to say "should these values be vars/constants somewhere instead of bare values", but `WebhookTimeout.Value("10")` is right there looking at me so I didn't :stuck_out_tongue: 
3. And now I have a 3rd question, why did only the admission plugin get this plumbing and not the recommender or the updater?

internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller_test.go
```diff
@@ -87,6 +87,27 @@ func includeString(list []string, item string) bool {
 	return false
 }
 
+func TestAdmissionArgs(t *testing.T) {
+	vpa := NewVerticalPodAutoscaler()
+
+	args := AdmissionPluginArgs(vpa, &Config{Namespace: TestNamespace})
+
+	expected := []string{
```
jkyros, line 95: Awesome! We've even got a test!

internal/controller/verticalpodautoscaler/admission_plugin.go
```diff
@@ -13,12 +15,20 @@ type AdmissionPluginArg string
...
 const (
+	KubeAPIQPSArg    AdmissionPluginArg = "--kube-api-qps"
+	KubeAPIBurstArg  AdmissionPluginArg = "--kube-api-burst"
 	TLSCertFileArg   AdmissionPluginArg = "--tls-cert-file"
 	TLSKeyFileArg    AdmissionPluginArg = "--tls-private-key"
 	TLSCACertFileArg AdmissionPluginArg = "--client-ca-file"
 	WebhookTimeout   AdmissionPluginArg = "--webhook-timeout-seconds"
 )
 
+func CheckArgExists(args []string, arg AdmissionPluginArg) bool {
```
jkyros, line 26: How do you feel about `ArgExists` instead of `CheckArgExists`?  And you could also probably stuff it in util, but I don't know that it's hurting anything living where it is.

jkyros (commented):
Thanks Max! I just have a couple of understanding questions.

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
maxcao13: Yeah, I can just remove it especially since we didn't cap it before, and I think I have to remove the limits on the rbac-proxy container too. That might happen if our operator is behaving badly for some reason and is over the limit.

internal/controller/verticalpodautoscaler/admission_plugin.go
maxcao13: 1. Yeah, I'm trying to set it as a default via the operator rather than through the operands.
2. :stuck_out_tongue: 
3. The defaults for those are already 5/10 for qps/burst and according to the spike/table anyways it should be fine, so I didn't touch them.

internal/controller/verticalpodautoscaler/admission_plugin.go
maxcao13: Sure sounds good, always open to better variable/function names :D

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
jkyros: Yeah, I wasn't implying it would sneak the limit in -- it would restart, but I just want to make sure we know we're not close. 

For history, Back when I was on the MCO, OpenShift turned on the FIPS compliance shim/builder thing, and the additional stuff it did to link the FIPS libraries put us over our memory limit. Customer MCOs started dying of OOM after an upgrade, and we were very confused for a little while because we didn't change anything, and we weren't leaking, but we were all of a sudden over. I think it was [OCPBUGS-17568](https://issues.redhat.com/browse/OCPBUGS-17568)

( MCO also had the problem though where our object size varied a lot, and people could feed us really fat manifests, whereas I don't think anyone is feeding us really fat VPA controllers ) 

Anyway, I just wanted to make sure there is enough slack so it's only protecting the customer's environment if we go haywire, and that we're not so close to the edge that if the libraries get fatter we start dying sometimes and upsetting them.  

I'd probably feel better about a 200+ mem limit, 500 would be just fine.  :smile:

joelsmith: I agree here. Let's change it to 500 MiB which gives a lot of head-room, but protects clusters against unforeseen memory leak bugs bringing a node to its knees.

config/manager/manager.yaml
```diff
@@ -66,6 +67,9 @@ spec:
           requests:
             cpu: 20m
             memory: 50Mi
+          limits:
+            cpu: 100m
+            memory: 250Mi
```
joelsmith, line 72: Seems like the limits here and in the bundle CSV don't match. Do we need another `make bundle` run (after setting the limit to 500Mi)?

joelsmith (commented):
Thanks for fixing this bug! I think this is ready to merge once the limits are updated.

config/manager/manager.yaml
maxcao13: Oh yeah, maybe. I'll change this to 500 and see if make bundle changes it. Thanks!

Makefile
```diff
@@ -290,6 +290,7 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.2.1
 CONTROLLER_TOOLS_VERSION ?= v0.14.0
+ENVTEST_VERSION ?= release-0.18
```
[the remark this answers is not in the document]
maxcao13: I think some CI tests were failing because of this so I changed this before the operator-sdk upgrade. But this should be superseded by https://github.com/openshift/vertical-pod-autoscaler-operator/pull/186 :)

config/manager/manager.yaml
maxcao13: Yup, just forgot :)

joelsmith (approved):
/lgtm

Thanks!

maxcao13:
No, I can't think of anything meaningful. I think it's a flake too, but let's see :/

PR #181, opened 2024-11-18
remove RELATED_* env var so it doesn't conflict with ART build

lgarciaaco:


lgarciaaco:
Closed in favor of https://github.com/openshift/vertical-pod-autoscaler-operator/pull/182

PR #182, opened 2024-11-21
remove RELATED_* env var so it doesn't conflict with ART build

lgarciaaco:


joelsmith:
/close
Closing in favor of #183

PR #183, opened 2024-11-26
OCPBUGS-45072: Rename operand image env var to satisfy ART build check

joelsmith:
Needed so we can revert https://github.com/openshift-eng/ocp-build-data/pull/5741

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -245,7 +245,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 predeploy: yq ## Setup configuration for the operator before deployment.
 	cd config/manager && $(KUSTOMIZE) edit set image quay.io/openshift/origin-vertical-pod-autoscaler-operator=$(OPERATOR_IMG)
 	$(YQ) eval 'del(.patches)' -i config/manager/kustomization.yaml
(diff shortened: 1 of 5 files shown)


maxcao13 (commented):
In the error in the JIRA card
```
   2024-11-22 20:30:04,900 artcommonlib INFO Task state: FREE
2024-11-22 20:33:04,945 artcommonlib INFO Fault: <Fault 2001: 'Image
build failed. Error in plugin pin_operator_digest: Both relatedImages
and RELATED_IMAGE_* env vars present in
/workspace/ws-build-dir/x86_64/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml.
Please remove the relatedImages section, it will be reconstructed
automatically.; . OSBS build id:
vertical-pod-autosca-rhaos-418-rhel-9-f331015954-20241122203110'>
```
It just states that we should remove the `relatedImages` section, which you merged 2 weeks ago. I assume there are just other errors that state that you can't have these types of `RELATED_IMAGE_*` environment variables anymore?

joelsmith:
> It just states that we should remove the `relatedImages` section, which you merged 2 weeks ago. I assume there are just other errors that state that you can't have these types of `RELATED_IMAGE_*` environment variables anymore?

I think ART is adding the related images section back in during the build, so they are detecting the env var presence in their generated CSV which has both

PR #186, opened 2024-11-27
OCPBUGS-45127: Upgrade to operator-sdk 1.38 to get rid of kube-rbac-proxy image

joelsmith:
We're currently failing the CVP FIPS test because of our use of the upstream image. As Max pointed out, the newest operator-sdk provides a pattern for performing metrics endpoint authentication and authorization without the use of the kube-rbac-proxy sidecar. This PR moves to operator-sdk 1.38.

Also, use the service name `vpa-operator-metrics` instead of `vpa-metrics-service` to make it clear that the metrics are operator metrics.

Followed `[go/v4] ` steps from:
* https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.36.0/
* https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.37.0/ (no-op)
* https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.38.0/

```
go get sigs.k8s.io/controller-runtime/pkg/metrics/filters@v0.18.4
go mod tidy
go mod vendor
```

Tested metrics endpoint authorization:
```
$ oc exec -ti -n openshift-vertical-pod-autoscaler deployments/vertical-pod-autoscaler-operator -- curl -k https://localhost:8443/metrics
Unauthorized
$ oc create serviceaccount test-sa
$ oc create -f - <<EOF
apiVersion: v1
kind: Secret
metadata:
  name: test-sa-token
  annotations:
    kubernetes.io/service-account.name: test-sa 
type: kubernetes.io/service-account-token
EOF
$ TOKEN=$(oc get secret test-sa-token -o jsonpath={.data.token} | base64 -d)
$ create clusterrolebinding test-sa --clusterrole=vpa-metrics-reader --serviceaccount=openshift-vertical-pod-autoscaler:test-sa
$ oc exec -ti -n openshift-vertical-pod-autoscaler deployments/vertical-pod-autoscaler-operator -- curl -k --header "Authorization: Bearer $TOKEN" https://localhost:8443/metrics
# HELP certwatcher_read_certificate_errors_total Total number of certificate read errors
# TYPE certwatcher_read_certificate_errors_total counter
certwatcher_read_certificate_errors_total 0
# HELP certwatcher_read_certificate_total Total number of certificate reads
# TYPE certwatcher_read_certificate_total counter
certwatcher_read_certificate_total 0
...
...
...
```


maxcao13 (commented):
Ah didn't think about that. Are we going to cherry-pick to 4.18 with the 4.18 version of the image as well?

joelsmith:
> Ah didn't think about that. Are we going to cherry-pick to 4.18 with the 4.18 version of the image as well?

Yes, I think we need to. We'll have to cherry-pick manually though since the cherry-pick bot would keep it at 4.18.

maxcao13 (commented):
Is the FIPS check okay with quay images but not images from gcr?

joelsmith:
> Is the FIPS check okay with quay images but not images from gcr?

The FIPS check will pass with most any image build with OpenShift's build process using OpenShift builder images. So it's more about switching to images which fit that mold.

> I think we could avoid this issue if we upgrade to operator-sdk 1.38? Seems like kubebuilder v4, which is what 1.38 upgrades to, deprecates the image entirely and integrates similar protection without using the separate image: https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.38.0/

That is a great idea, thank you for bringing it up! I think we can fix this bug via that approach and be in a much better spot.

Makefile
```diff
@@ -286,32 +278,49 @@ KUBECTL ?= kubectl
 KUSTOMIZE ?= $(LOCALBIN)/kustomize
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
 ENVTEST ?= $(LOCALBIN)/setup-envtest
+GOLANGCI_LINT = $(LOCALBIN)/golangci-lint
 
 ## Tool Versions
-KUSTOMIZE_VERSION ?= v5.2.1
-CONTROLLER_TOOLS_VERSION ?= v0.14.0
+KUSTOMIZE_VERSION ?= v5.3.0
+KUSTOMIZE_VERSION ?= v5.4.2
```
maxcao13, line 285: I think you duplicated this?

Makefile
```diff
@@ -286,32 +278,49 @@ KUBECTL ?= kubectl
...
+CONTROLLER_TOOLS_VERSION ?= v0.15.0
+ENVTEST_VERSION ?= release-0.18
+GOLANGCI_LINT_VERSION ?= v1.59.1
 
-# GOFLAGS is explicitly empty here so that we do not run into https://github.com/golang/go/issues/45811 on CI
```
maxcao13, line 294: Should we keep this comment for future maintainability? I'm assuming you re-added the GOFLAGs='' in the "go-install-tool" thing because it failed in CI...?

Makefile
joelsmith: Oops, good catch

Makefile
joelsmith: Good call

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
```diff
@@ -759,6 +740,42 @@ spec:
...
+          - coordination.k8s.io
+          resourceNames:
+          - vpa-updater
+          resources:
+          - leases
+          verbs:
+          - get
+          - watch
+          - update
+        serviceAccountName: vpa-updater
```
maxcao13, line 778: Do you know where these rules came from?

cmd/main.go
```diff
@@ -98,7 +102,30 @@ func main() {
...
+		BindAddress:   metricsAddr,
+		SecureServing: secureMetrics,
+		// TODO(user): TLSOpts is used to allow configuring the TLS config used for the server. If certificates are
+		// not provided, self-signed certificates will be generated by default. This option is not recommended for
+		// production environments as self-signed certificates do not offer the same level of trust and security
+		// as certificates issued by a trusted Certificate Authority (CA). The primary risk is potentially allowing
+		// unauthorized access to sensitive metrics data. Consider replacing with CertDir, CertName, and KeyName
+		// to provide certificates, ensuring the server communicates using trusted and secure certificates.
+		TLSOpts: tlsOpts,
+	}
```
maxcao13, line 119: Is it okay for us to use self-signed certs for now? Are there potential risks if the vpa-operator's metrics were ever exposed?

joelsmith: I don't think that there is anything particularly sensitive in the metrics. The token itself (used to authenticate against the metrics endpoint) is potentially sensitive as it might be more privileged than just a token from a single-use service account, so it's probably a good idea to do this right.

At the same time, doing it right on a non-OpenShift cluster is rather difficult. What do you think of this idea:
- We annotate the service object such that OpenShift will create a service cert for us
- We mount the CM & secrets that OpenShift should create, but we make the mount optional so that the pod can start even if running on a non-OpenShift cluster
- We check to see if the cert & key exist at runtime in the operator, and if so, we pass them in, and if not, we set the options to generate a temporary self-signed cert

That seems like a reasonable compromise:
- On OpenShift, it just works securely out of the box
- On non-OpenShift (if anybody is using it there), it works out of the box, but less securely
- On non-OpenShift, someone can make it fully secure by creating a legit CA-signed cert and placing them where OpenShift would have placed them, and we'd start using their cert.

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
joelsmith: Yes, that reveals a process problem in this repo. They came from #180, but I didn't run `make bundle` on that PR. We should probably have a CI step which runs `make bundle` and compares to the current bundle to the CI-generated one to make sure there are no real differences and catches me when I forget to run it.

maxcao13: I guess it's from here but we didn't do `make bundle` somewhere. https://github.com/openshift/vertical-pod-autoscaler-operator/blob/8e649b20074cb06f0d3839c27f2be9507d6f1011/config/vpa/vpa-rbac.yaml#L370

cmd/main.go
maxcao13: Sure that sounds reasonable to me, and your explanation makes sense. service-ca operator makes this simpler. Sorry for opening pandora's box with suggesting this upgrade this late :stuck_out_tongue:

joelsmith:
Okay, I have it working with the automatically-issued TLS service certs. Those changes are in the top commit "Add trusted TLS cert to operator metrics endpoint"

```
$ oc project openshift-vertical-pod-autoscaler
$ oc create serviceaccount test-sa
$ oc create -f - <<EOF
apiVersion: v1
kind: Secret
metadata:
  name: test-sa-token
  annotations:
    kubernetes.io/service-account.name: test-sa 
type: kubernetes.io/service-account-token
EOF
$ oc create clusterrolebinding test-sa --clusterrole=vpa-metrics-reader --serviceaccount=openshift-vertical-pod-autoscaler:test-sa
$ TOKEN=$(oc get secret test-sa-token -o jsonpath={.data.token} | base64 -d)
$ oc port-forward service/vpa-operator-metrics 8443:https &
$ curl --cacert <(oc get cm openshift-service-ca.crt -o jsonpath="{.data['service-ca\.crt']}")  --header "Authorization: Bearer $TOKEN" --resolve vpa-operator-metrics.openshift-vertical-pod-autoscaler.svc:8443:127.0.0.1 https://vpa-operator-metrics.openshift-vertical-pod-autoscaler.svc:8443/metrics
# HELP certwatcher_read_certificate_errors_total Total number of certificate read errors
# TYPE certwatcher_read_certificate_errors_total counter
certwatcher_read_certificate_errors_total 0
# HELP certwatcher_read_certificate_total Total number of certificate reads
# TYPE certwatcher_read_certificate_total counter
certwatcher_read_certificate_total 1

...
...
...
$ kill %1
```

config/rbac/metrics_reader_role.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: metrics-reader
```
maxcao13, line 4:
```
$ oc project openshift-vertical-pod-autoscaler
$ oc create serviceaccount test-sa
$ oc create -f - <<EOF
apiVersion: v1
kind: Secret
metadata:
  name: test-sa-token
  annotations:
    kubernetes.io/service-account.name: test-sa 
type: kubernetes.io/service-account-token
EOF
$ oc create clusterrolebinding test-sa --clusterrole=vpa-metrics-reader --serviceaccount=openshift-vertical-pod-autoscaler:test-sa
...
...
```
In your example, you used the `vpa-metrics-reader` cluster role. Should we rename this to that?

maxcao13:
It wasn't in the upgrade guide (I'm not sure why), but I think we should bump this up too: https://github.com/openshift/vertical-pod-autoscaler-operator/pull/186/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L60

joelsmith:
> It wasn't in the upgrade guide (I'm not sure why), but I think we should bump this up too: https://github.com/openshift/vertical-pod-autoscaler-operator/pull/186/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L60

Good catch
 
> And after that, I think if you delete the `vertical-pod-autoscaler-operator/bin` directory, and run `make bundle` again, it should regenerate operator-sdk to version 1.38 and update the csv again as well.
> 
> ```yaml
>     operators.operatorframework.io/builder: operator-sdk-v1.35.0 -> 1.38.0
> ```

The makefile also supports using `operator-sdk` from the `PATH` and I already had 1.38 for the work I've been doing on this. `make bundle` did not change that as it comes from here: https://github.com/openshift/vertical-pod-autoscaler-operator/blob/795172b1e0b1ec71ee31c58b8f6e802ec9624b8b/config/manifests/bases/vertical-pod-autoscaler.clusterserviceversion.yaml#L44

config/rbac/metrics_reader_role.yaml
joelsmith: Hmm, looks like we have a lot of old cruft in the bundle dir, which is where `vpa-metrics-reader` was coming from. Maybe we need to delete `bundle/manifests/` and re-generate it as part of the new test, to detect manifests which are out-of-date relative to the config.

And separate from that, I should have been more deliberate in following the upgrade steps instead of just wholesale copying what they put. Some of the names should be our own, but some of them, I'm not sure what we should do. The `metrics-reader` and `system:metrics-reader` cluster roles are not VPA specific. I'm not sure why `operator-sdk` is having us add them to our manifests (and probably doing the same to everyone else's manifests too). Maybe it's like a common dependency and any and all installed operators will provide them if they're not already there.

config/rbac/metrics_auth_role_binding.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: metrics-auth-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: metrics-auth-role
+subjects:
+- kind: ServiceAccount
+  name: vertical-pod-autoscaler-operator
+  namespace: system
```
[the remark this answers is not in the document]
joelsmith: Oops, I missed that one. Once I changed it, the diff size of the CSV went down quite a bit.

Yep, the RBAC is to create tokenreviews (ask the API server if a token is valid -- authentication) and subjectaccessreviews (ask the API server if the subject has rights to a resource -- authorization). Whether the kube-rbac-proxy is doing that job, or the operator is doing that job directly without the proxy, they both would need that access to be able to ask the API server about the authn/authz of the metrics request they're receiving.

config/rbac/metrics_reader_role.yaml
joelsmith: It looks like `system:metrics-reader` comes from upstream VPA RBAC which we bring in https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/deploy/vpa-rbac.yaml#L1-L13

and `metrics-reader` comes from the final step of https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.38.0/

So in both cases, I think they are intended to be globally useful roles. The former for getting pod metrics (useful for HPA, VPA, etc.) and the latter for reaching metrics endpoints (for Prometheus, et al. to be able to scrape all metrics targets in the cluster with a single rolebinding).

I think we leave those two as-is.

maxcao13: Oh, ok makes sense, but confused me on the naming. Thanks for getting to the bottom of that.

joelsmith: Yeah, I'm glad you brought it up since it helped me realize about the stale bundle issue so we could get that cleaned up.

maxcao13:
> The makefile also supports using operator-sdk from the PATH and I already had 1.38 for the work I've been doing on this. make bundle did not change that as it comes from here: 
Hmm... yeah I guess that annotation comes from whatever you initialize the project with. I hate that there's very little documentation for a lot of the annotations in there, and what they do.

/lgtm

PR #194, opened 2024-11-29
Remove replaces from manifest

joepvd:
OCP puts in `skips`. There is no need for `replaces`, and this make a release pipeline hurt, see e.g. https://iib.engineering.redhat.com/api/v1/builds/874123/logs


maxcao13 (commented):
Thanks for fixing that!

/lgtm
/approve

PR #197, opened 2024-12-04
PODAUTO-284: Add CI checks to ensure "make bundle/generate" before pushing 

maxcao13:
Add ensure-commands-are-noops ci check such that we make sure that a dev has run make bundle and make generate before pushing their changes.

~Also removes gofmt and govet make targets since golangci-lint already has rules that do the same thing~, and fixes up `make check` such that it includes all commands that devs can and should use before pushing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,7 +158,12 @@ test-scorecard: operator-sdk ## Run the scorecard tests. Requires an OpenShift c
 	$(OPERATOR_SDK) scorecard bundle -n default -w 300s
 
 .PHONY: check
-check: fmt vet manifest-diff lint test ## Check code for formatting, vet, lint, manifest-diff and run tests.
+check: manifest-diff lint yamllint test ## Run quick checks for a dev before pushing code.
+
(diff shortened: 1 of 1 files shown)


maxcao13:
I nabbed code from here: https://github.com/jaegertracing/jaeger-operator/blob/3ebc5b4431d74917c4d1bfe074b4faeea319e02d/Makefile#L122-L127 and added some git diff options:
* `-I` - [ignore matching lines](https://www.sjoerdlangkemper.nl/2021/08/13/how-does-git-diff-ignore-matching-lines-work/) to ignore the `createdAt` line in the csv that changes everytime: 
* `-P` - which tells Git not to use a pager (which I don't know matters for CI but shouldn't hurt)

maxcao13:
I'm going re-add the make targets I removed, because I think it's okay to have them individually before running `make test` but I'll remove the CI steps in the release repo.

joelsmith:
Nice! One less thing I'll forget to do before merging my PRs!

/lgtm

PR #199, opened 2024-12-06
Clean old bundles before regenerating

joelsmith:
This is a follow-up to #197

Un-owned files from previous versions were being left around in the bundle without manually removing the bundle/manifests dir prior to regenerating.


maxcao13:
/lgtm

Thanks!

PR #206, opened 2025-06-30
OCPBUGS-57966: Updates for 4.20

maxcao13:
```
sed -i 's/4.19/4.20/g' $(git grep -l 4.19 config/) hack/manifest-diff-upstream.sh hack/e2e.sh Makefile bundle/image-references bundle/vertical-pod-autoscaler-operator.package.yaml
```

This commit updates references in the repo from 4.19 to 4.20, except for actually bumping go dependencies. We need this to satisfy ART builds, but we cannot do a full update for our OpenShift 4.20 dependencices yet due to the k8s 1.33 rebase not being finished, so we will hold off on golang dep bumps, go1.24 bumps, and image bumps until then.

We will have a separate PR for that once the rebase resolves.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 # Must be semver compliant
-export OPERATOR_VERSION ?= 4.19.0
+export OPERATOR_VERSION ?= 4.20.0
 OPERATOR_PKG_NAME ?= vertical-pod-autoscaler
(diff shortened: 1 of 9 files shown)


jkyros:
Makes sense. Unfortunate how the "chore cycle" lines up sometimes :cry: Thanks Max! 
/override ci/prow/manifest-diff
/lgtm

PR #207, opened 2025-07-01
AUTOSCALE-283: Annotation logic for controller creation and test additions

fishereskew:
Added logic to annotate namespace to prevent another default controller from being created. Annotation logic was implemented to prevent a scenario similar to that in: https://github.com/kedacore/keda-olm-operator/pull/255

**verticalpodautoscaler_controller.go:**
- Changed logic to mimic that of the code in the keda-olm-operator to prevent duplicate annotation code and so that it uses WaitForCacheSync() instead of relying on 60 attempts with 1-second intervals.

**e2e.sh:**
- Added function to test the controller using the annotation to prevent another default controller from being created.

**autoscaling_v1_verticalpodautoscalercontroller.yaml:**
- Added required namespace for the controller to be recognized.
- _Let me know if this was left without the namespace for a reason I'm not aware of_

**role.yaml:**
- Added namespaces for controller code to be able to access the operator namespace.

**README.md & Makefile:**
- A cluster with the operator running is required for the `test-e2e` command to work properly so I changed the comments to try to reflect that and be a bit more clear.


fishereskew:
Question:

Is there a reason why autoscaling_v1_verticalpodautoscalercontroller.yaml doesn't have a namespace? When I run `make create_vpa_controller_cr`, it doesn't start up the 3 controllers because the controller goes to the default namespace instead of the openshift-vertical-pod-autoscaler namespace. Is the namespace empty for the user to somehow define what namespace it should be in or will it always be openshift-vertical-pod-autoscaler?

config/samples/autoscaling_v1_verticalpodautoscalercontroller.yaml
fishereskew, line 1: Is there a reason why autoscaling_v1_verticalpodautoscalercontroller.yaml doesn't have a namespace? When I run make create_vpa_controller_cr, it doesn't start up the 3 controllers because the controller goes to the default namespace instead of the openshift-vertical-pod-autoscaler namespace. Is the namespace empty for the user to somehow define what namespace it should be in or will it always be openshift-vertical-pod-autoscaler?

fishereskew:
Update: added ginko+gomega test suite using envTest. Did a basic ginkgo test translation for other test files.

config/testing/vpa_namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-vertical-pod-autoscaler
```
maxcao13, line 4: I'd rather not a create this file just to add a namespace to the test environment. Can we just create it  directly somehow in the setup, or is that not possible?

internal/controller/verticalpodautoscaler/suite_test.go
```diff
@@ -0,0 +1,108 @@
...
+	var err error
+
+	err = vpav1.AddToScheme(scheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	// +kubebuilder:scaffold:scheme
+
+	By("bootstrapping test environment")
+
+	// cfg, testEnv, and k8sClient
```
maxcao13, line 76: I think it's best to we don't write comments like this. It's not really telling us anything that the code doesn't already.

internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
```diff
@@ -53,6 +53,9 @@ import (
 const (
 	// ControllerName The hard-coded name of the VPA controller
 	ControllerName = "vertical-pod-autoscaler-controller"
+	// Operator namespace annotation for default controller creation
+	DefaultVPAControllerAnnotation = "vertical-pod-autoscaler-controller/created-default-controller"
+	AnnotationTrue                 = "true"
```
maxcao13, line 58: I think "true" without declaring our own variable is just fine here. (Like we are anticipating that the definition of "true" is going to change in the future of Kubernetes, which I feel is unlikely).

go.mod
```diff
@@ -7,14 +7,19 @@ toolchain go1.23.4
 require (
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/go-logr/logr v1.4.2
+	github.com/manifestival/controller-runtime-client v0.4.0
```
maxcao13, line 10: What's the reason for bumping these deps? Generally we probably shouldn't do it unless it's necessary, or part of release chores, and if we need to, it's best to put it in a separate commit to keep it contained and easier to review and view past history..

internal/util/util_test.go
```diff
@@ -1,280 +1,266 @@
 package util
```
maxcao13, line 2: This is sort of hard to review. Can we separate the refactoring of all the unit tests to ginkgo/gomega in separate commits or a separate PR?

config/samples/autoscaling_v1_verticalpodautoscalercontroller.yaml
fishereskew: Doing it via makefile seems much better

internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
fishereskew: That is only there because lint was failing for multiple uses of the string "true"

fishereskew: I could change the value from "true" to something else if that would be better

maxcao13: Right, since annotations can be whatever we want, that makes sense. I was thinking that it was the boolean true.

Ignore me then, it's probably fine as it is. :stuck_out_tongue:

internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller_unit_test.go
```diff
@@ -0,0 +1,399 @@
...
+				case "vpa-recommender":
+					Expect(podSpec.Containers[0].Resources).To(Equal(vpa.Spec.DeploymentOverrides.Recommender.Container.Resources))
+				case "vpa-updater":
+					Expect(podSpec.Containers[0].Resources).To(Equal(vpa.Spec.DeploymentOverrides.Updater.Container.Resources))
+
+				}
+			}
+		})
+
+		It("should overide args", func() {
```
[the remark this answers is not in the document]
fishereskew: lol, thanks

internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller_unit_test.go
```diff
@@ -0,0 +1,399 @@
...
+
+		Expect(args).To(ContainElement(fmt.Sprintf("--recommendation-margin-fraction=%.01f", RecommendationMarginFraction)))
+		Expect(args).To(ContainElement(fmt.Sprintf("--pod-recommendation-min-cpu-millicores=%.01f", PodRecommendationMinCPUMilicores)))
+		Expect(args).To(ContainElement(fmt.Sprintf("--pod-recommendation-min-memory-mb=%.0f", PodRecommendationMinMemoryMB)))
+
+		Expect(args).NotTo(ContainElement("--scale-down-delay-after-delete"))
+		Expect(args).NotTo(ContainElement("--scale-down-delay-after-failure"))
+	})
+
+	Describe("When handeling overrides", func() {
```
maxcao13, line 115: `handling`

fishereskew: Alright now this is just sad...

internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller_unit_test.go
```diff
@@ -0,0 +1,399 @@
...
+				expectedRes: reconcile.Result{},
+				c:           &cfg2,
+				d:           &dep1,
+			},
+			// Case 2: no dep found, should pass, returns {}, nil.
+			{
+				expectedRes: reconcile.Result{},
+				c:           &cfg1,
+				d:           &appsv1.Deployment{},
+			},
```
maxcao13, line 263: I feel like we should be more explicit in the naming here. Right now each test case run looks like this:
```
VerticalPodAutoscalerController Unit Tests Reconciler Function should deploy each case 0
/home/macao/go/src/github.com/openshift/vertical-pod-autoscaler-operator/internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller_unit_test.go:267
I0805 12:46:39.630047   62734 verticalpodautoscaler_controller.go:452] Creating VerticalPodAutoscalerController deployment: test-namespace/vpa-recommender-test
I0805 12:46:39.630127   62734 verticalpodautoscaler_controller.go:239] Created VerticalPodAutoscalerController deployment: test-namespace/vpa-recommender-test
I0805 12:46:39.630140   62734 verticalpodautoscaler_controller.go:452] Creating VerticalPodAutoscalerController deployment: test-namespace/vpa-updater-test
I0805 12:46:39.630173   62734 verticalpodautoscaler_controller.go:239] Created VerticalPodAutoscalerController deployment: test-namespace/vpa-updater-test
I0805 12:46:39.630184   62734 verticalpodautoscaler_controller.go:452] Creating VerticalPodAutoscalerController deployment: test-namespace/vpa-admission-plugin-test
I0805 12:46:39.630229   62734 verticalpodautoscaler_controller.go:239] Created VerticalPodAutoscalerController deployment: test-namespace/vpa-admission-plugin-test
I0805 12:46:39.630244   62734 verticalpodautoscaler_controller.go:499] Creating VerticalPodAutoscalerController service: vpa-webhook
I0805 12:46:39.630276   62734 verticalpodautoscaler_controller.go:281] Created VerticalPodAutoscalerController service: vpa-webhook
I0805 12:46:39.630291   62734 verticalpodautoscaler_controller.go:541] Creating VerticalPodAutoscalerController configmap: vpa-tls-ca-certs
I0805 12:46:39.630315   62734 verticalpodautoscaler_controller.go:321] Created VerticalPodAutoscalerController ConfigMap: vpa-tls-ca-certs
• [0.001 seconds]
```
If we were looking at this through the CI test run viewer, I think it would be a lot better to have a general idea of what the case was, rather than going to the code and then cross referencing it to these comments here.

Same with all the other test case names of the same ilk.

fishereskew:
```
tCases := []struct {
			label       string
			expectedRes reconcile.Result
			c           *Config
			d           *appsv1.Deployment
		}{
			{
				label:       "vpa and dep found, returns {}, nil",
				expectedRes: reconcile.Result{},
				c:           &cfg1,
				d:           &dep1,
			},
			{
				label:       "no vpa found, returns {}, nil",
				expectedRes: reconcile.Result{},
				c:           &cfg2,
				d:           &dep1,
			},
			{
				label:       "no dep found, returns {}, nil",
				expectedRes: reconcile.Result{},
				c:           &cfg1,
				d:           &appsv1.Deployment{},
			},
		}

		for _, tc := range tCases {
			It(fmt.Sprintf("should deploy when: %s", tc.label), func() {
```
			

Was thinking to just add a label for all of them if that's sufficient

maxcao13: Makes sense to me :)

maxcao13:
I'm running the operator and I see a lot of these in the operator logs:
```go
E0805 21:33:38.666717 1 reflector.go:166] "Unhandled Error" err="sigs.k8s.io/controller-runtime/pkg/cache/internal/informers.go:108: Failed to watch *v1.Namespace: namespaces is forbidden: User \"system:serviceaccount:openshift-vertical-pod-autoscaler:vertical-pod-autoscaler-operator\" cannot watch resource \"namespaces\" in API group \"\" at the cluster scope" logger="UnhandledError"
```
I think we need to add a `watch` rule for namespaces? Although I don't really know why, becasue we are not setting up explicit watches for namespaces. I'm assuming operator-sdk code is doing something weird here.

maxcao13:
```go
0805 21:20:04.963068 1 verticalpodautoscaler_controller.go:232] Error creating VerticalPodAutoscalerController deployment: deployments.apps "vpa-recommender-default" is forbidden: cannot set blockOwnerDeletion in this case because cannot find RESTMapping for APIVersion autoscaling.openshift.io/v1 Kind VerticalPodAutoscalerController: no matches for kind "VerticalPodAutoscalerController" in version "autoscaling.openshift.io/v1"
2025-08-05T21:20:04Z ERROR Reconciler error {"controller": "verticalpodautoscalercontroller", "controllerGroup": "autoscaling.openshift.io", "controllerKind": "VerticalPodAutoscalerController", "VerticalPodAutoscalerController": {"name":"default","namespace":"openshift-vertical-pod-autoscaler"}, "namespace": "openshift-vertical-pod-autoscaler", "name": "default", "reconcileID": "1187c43f-d046-4ef1-acc9-08ea349e1fbb", "error": "deployments.apps \"vpa-recommender-default\" is forbidden: cannot set blockOwnerDeletion in this case because cannot find RESTMapping for APIVersion autoscaling.openshift.io/v1 Kind VerticalPodAutoscalerController: no matches for kind \"VerticalPodAutoscalerController\" in version \"autoscaling.openshift.io/v1\""}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:263
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:224
```
I also see these errors on startup, but they go away quickly after, and the startup proceeds.

Not really sure why that is. I put in a 20 second sleep like this:
```go
// SetupWithManager sets up the controller with the Manager.
func (r *VerticalPodAutoscalerControllerReconciler) SetupWithManager(mgr ctrl.Manager) error {
       time.Sleep(20)
	// Create a default VPAController upon first operator startup
	go func() {
		err := r.ensureVPAController(mgr)
...
```
And the error stopped showing up, so something in there is too fast? Really weird.

fishereskew:
@maxcao13 not really sure why that's happening. I fixed the watch rbac issue and now I'm getting a reconciler error for configmaps already existing when it tries to patch the annotation:

```
ERROR	Reconciler error	{"controller": "verticalpodautoscalercontroller", "controllerGroup": "autoscaling.openshift.io", "controllerKind": "VerticalPodAutoscalerController", "VerticalPodAutoscalerController": {"name":"default","namespace":"openshift-vertical-pod-autoscaler"}, "namespace": "openshift-vertical-pod-autoscaler", "name": "default", "reconcileID": "628d8aa0-6111-47c6-a0b4-0f5a5785c7a5", "error": "configmaps \"vpa-tls-ca-certs\" already exists"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:263
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
	/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:224
```

Sleeping for 20 seconds also makes this error go away.

PR #208, opened 2025-07-23
Update CRDs & RBAC to match operand

joelsmith:
Copy data from `https://raw.githubusercontent.com/openshift/kubernetes-autoscaler/release-4.20/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml` to `config/vpa/vpa-crd.yaml` and `config/vpa/vpacheckpoint-crd.yaml`

Also update `config/vpa/vpa-rbac.yaml`


joelsmith:
This is also blocking #207 since it causes blocking CI tests to fail

PR #209, opened 2025-09-24
OCPBUGS-61456: Reduce number of version bumps for release chores

maxcao13:
* [fix: cleanup image references to not reference versions](https://github.com/openshift/vertical-pod-autoscaler-operator/commit/39975bd3b9048a2f59009cbe2411afed0805e911)
  * Removes versioning in the image-references art file so we don't have to maintain that it has the latest major.minor.z tag.
  * I believe it is safe to do this since we don't do anything with it on the dev side. And ART is simply just using it for search and replace according to their docs.
* [chore(art): replace versioning for ART package.yaml file with a place](https://github.com/openshift/vertical-pod-autoscaler-operator/commit/e535cbd94153f90db45dc8cfc792746dc2de7ebd)
  * similar idea to the first commit, I don't think we need to maintain a versioning for this, and since ART uses a search and replace thingy for their substitution.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 # This variable is used to construct full image tags for bundle and catalog images.
 #
 # For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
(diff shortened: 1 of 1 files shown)


maxcao13:
I think I'll just close this. It's not a huge deal to get this in, we have a long history of precedent of what to bump anyways.

EDIT: Although this commit changes are still worthwhile: https://github.com/openshift/vertical-pod-autoscaler-operator/pull/209/commits/500c0389676dd31dc8f1020c0b60ab3b94d155d6 I'll probably shoehorn that into some other PR in the future.

PR #211, opened 2025-10-01
OCPBUGS-60059: Fix version injection regression

jkyros:
We broke version injection with #169, this just re-adds version injection so we stop getting the "improperly built" version errors. 

It does this by: 
- Adding some version injection vars in the Makefile 
- Copying the Makefile and .git directory into the docker build stages so the version can be retrieved (we used to include the git hash of the build in our version before we broke it) 
- Adding a simple binary build target in the Makefile and calling it from the dockerfile 
- I don't think I broke multi-arch

**NOTE:** We used to copy everything `COPY . . ` so we can go back to that if that's just easier, but I was assuming we found value in explicitly listing just the things in the Dockerfile that the build needs to live?

Now we get: 
```
jkyros@jkyros-thinkpadp16vgen1:~/dev/vertical-pod-autoscaler-operator$ podman run  localhost/vpa:latest
2025-10-01T22:54:55Z    INFO    setup   Go Version      {"version": "go1.23.9 (Red Hat 1.23.9-1.el9_6)"}
2025-10-01T22:54:55Z    INFO    setup   Go OS/Arch      {"os": "linux", "arch": "amd64"}
2025-10-01T22:54:55Z    INFO    setup   Version {"version": "vertical-pod-autoscaler-operator v4.20.0-fcb0d8f"}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,12 @@ BUNDLE_VERSION ?= $(IMAGE_VERSION)
 BUNDLE_MANIFESTS_DIR ?= $(shell pwd)/bundle/manifests
 OLM_MANIFESTS_DIR ?= $(shell pwd)/config/olm-catalog
 
+# We used to inject the git hash for version before https://github.com/openshift/vertical-pod-autoscaler-operator/pull/169, accidentally
+# removed it. This is for parity with the old behavior, should give us e.g. "v4.21.0-2332559":
+INJECT_VERSION     ?= v${OPERATOR_VERSION}-$(shell git rev-parse --short=7 HEAD)
+# version used to be in pkg, but now it's internal 
+LD_FLAGS    ?= -X github.com/openshift/vertical-pod-autoscaler-operator/internal/version.Raw=$(INJECT_VERSION)
+
 OUTPUT_DIR := ./_output
 OLM_OUTPUT_DIR := $(OUTPUT_DIR)/olm-catalog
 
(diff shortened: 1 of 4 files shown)


jkyros:
I'm not sure I understand yet how: 
- 4.20 operand manifests were last updated 5 months ago 
- our operand manifests here were last updated 3 months ago in #208 
- I didn't touch manifests at all with the version thing
- somehow manifest-diff still thinks we're askew? :melting_face:

joelsmith:
/lgtm

Do we need to override the tests? Seems like we're failing due to in-place not working, but I didn't look very deeply yet

maxcao13:
> Do we need to override the tests? Seems like we're failing due to in-place not working, but I didn't look very deeply yet
Seems reasonable to me, I don't have an opinion either way. Either we get this fixed sooner than later.

Makefile
```diff
@@ -190,12 +196,16 @@ e2e-olm-local: full-olm-deploy test-e2e
...
 .PHONY: build
 build: manifests generate fmt vet ## Build manager binary.
-	go build -o bin/manager cmd/main.go
+	go build -ldflags "$(LD_FLAGS)" -o bin/manager cmd/main.go
 
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./cmd/main.go
 
+.PHONY: conaeiner-binary-build
```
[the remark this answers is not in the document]
jkyros: hahaha well that was egregious, whoops. Great catch. _It's 2025, @jkyros, how do you not have a pre-PR job that runs a spell check?_ 

No comment.

jkyros:
RE: scorecard, I moved us to [aws-2 in July](https://github.com/openshift/release/pull/67521) because `aws`  was this broken, and now `aws-2` is doing the same thing :melting_face:

joelsmith:
I think that the tests should pass now once you rebase.

hack/e2e.sh
```diff
@@ -219,6 +269,18 @@ if ! wait_for_controllers 10; then
...
+echo "================================"
+echo "Verifying Operator Version"
+echo "================================"
+
+if ! verify_operator_version; then
+  echo ""
+  echo "ERROR: Operator version verification failed"
+  exit 1
+fi
+
```
maxcao13: We have to move this before `disable_controllers`, because that step disables the operator for admission-controller and updater tests :sweat_smile:. Full vpa suite doesn't do the operator disabling at all, so that's why the existing e2e didn't catch it.

I guess we should wait for the operator first, and then verify operator version, and then `disable_controllers`?

jkyros: oh the conditionality is _inside_ the disable function, that's kinda cursed :stuck_out_tongue: 

Fixed it so the version check happens before we shut the operator off. Thanks!

maxcao13:
This is a nit, so if it's not worth it then just ignore me:
```
Extracted version: v4.21.0-6cfcb9a"}
Operator version is valid: v4.21.0-6cfcb9a"}
```
The regex is erroneously matching the `"}` as well

Otherwise, I like that we've been building operator validations in this script :stuck_out_tongue:   

/lgtm

maxcao13:
/lgtm

Thanks!

PR #213, opened 2025-11-03
AUTOSCALE-433: allow more VPA suites to be tested

maxcao13:
This commit improves the hack/e2e.sh script and allows it to exercise more VPA suites. These include the admission-controller, recommender, updater, and actuation suites.

This needs to be followed up by some openshift/release PRs which will actually run these tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,10 +159,10 @@ ensure-commands-are-noops: generate bundle
 
 ##@ E2E Tests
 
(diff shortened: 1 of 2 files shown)


joelsmith:
> I can just add that small step back in there (don't manual scaling everything else, only do this through the VPAController object), during the recommender component test, would that work?

Yes, that sounds totally workable. Thanks!

PR #215, opened 2025-11-11
AUTOSCALE-433: Fixup e2e.sh hack script more

maxcao13:
The PR has two parts:

* The default ginkgo timeout is 1 hour. I am overriding it to 2 hours in the e2e hack script.
* I also extend the regular go test timeout to 125 minutes so that if by whatever reason somehow the ginkgo timeout doesn't happen at 2 hours, the go test timeout will still panic and stop the test eventually.
* Apply VPA_NAMESPACE env variable to script for custom namespace for tests

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -57,10 +57,11 @@ function run_upstream_vpa_tests() {
   echo "Running ${SUITE} e2e tests from upstream..."
   pushd "${SCRIPT_ROOT}/e2e" > /dev/null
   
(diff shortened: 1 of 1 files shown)


joelsmith:
Increasing the timeouts seems very smart. I really really hope we don't start seeing 2-hour runtimes by default!
Please make sure that https://github.com/openshift/kubernetes-autoscaler/pull/394 is merged and the tests here are re-run before adding the verified label.
/lgtm

maxcao13:
/verified by e2e-aws-olm

/cc @prozehna
for visibility

PR #217, opened 2025-11-21
Remove deprecated critical-pod annotation

### LINKED ISSUE #216: scheduler.alpha.kubernetes.io/critical-pod non-functional in v1.16+; use the "priorityClassName" field instead
https://github.com/openshift/vertical-pod-autoscaler-operator/blob/53df89ae079604ed50a012596593d7b6bfb09c9c/internal/util/util.go#L16

Saw this log line on the operator and want to make sure we've updated to get the desired behavior:

```
2025-11-20T15:39:27Z INFO KubeAPIWarningLogger spec.template.metadata.annotations[scheduler.alpha.kubernetes.io/critical-pod]: non-functional in v1.16+; use the "priorityClassName" field instead
```

cuppett:
## Summary

Fixes #216

This PR removes all references to the deprecated `scheduler.alpha.kubernetes.io/critical-pod` annotation that has been non-functional since Kubernetes v1.16+.

The operator already correctly uses `priorityClassName: system-cluster-critical` for VPA components, which is the proper Kubernetes-native way to mark pods as critical. This change eliminates the deprecation warning that appears in operator logs.

## Changes

- Removed `CriticalPodAnnotation` constant from `internal/util/util.go`
- Removed annotation usage from `UpdateAnnotations()` method in `internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go:633`
- Removed annotation usage from `AutoscalerDeployment()` method in `internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go:680`
- Updated test expectations in `TestUpdateAnnotations` to no longer expect the deprecated annotation

## Test plan

- [x] All existing tests pass (`make test`)
- [x] No references to the deprecated annotation remain in the codebase
- [x] The operator continues to properly set `priorityClassName` for critical pods

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller_test.go
+++ b/internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller_test.go
@@ -406,23 +406,36 @@ func TestUpdateAnnotations(t *testing.T) {
 		},
 	})
 
-	expected := map[string]string{
-		util.ReleaseVersionAnnotation: TestReleaseVersion,
-	}
-
 	testCases := []struct {
-		label  string
-		object metav1.Object
+		label    string
+		object   metav1.Object
+		expected map[string]string
 	}{
 		{
 			label:  "no prior annotations",
 			object: deployment.Object(),
+			expected: map[string]string{
+				util.ReleaseVersionAnnotation: TestReleaseVersion,
+			},
+		},
+		{
+			label: "missing version annotation",
+			object: deployment.WithAnnotations(map[string]string{
+				"some.other/annotation": "value",
+			}).Object(),
+			expected: map[string]string{
+				util.ReleaseVersionAnnotation: TestReleaseVersion,
+				"some.other/annotation":       "value",
+			},
 		},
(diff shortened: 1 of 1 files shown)


internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller_test.go
```diff
@@ -419,18 +418,6 @@ func TestUpdateAnnotations(t *testing.T) {
...
-			object: deployment.WithAnnotations(map[string]string{
-				util.CriticalPodAnnotation: "",
-			}).Object(),
-		},
-		{
-			label: "missing critical-pod annotation",
-			object: deployment.WithAnnotations(map[string]string{
-				util.ReleaseVersionAnnotation: TestReleaseVersion,
-			}).Object(),
-		},
```
jkyros: Claude was just like "These test are cross-wired, but since they both mention critical pod, I'll remove them both!" :laughing:

PR #228, opened 2026-05-19
AUTOSCALE-563: Add support for centralized TLS profile

joelsmith:
Also:
* Enable secure metrics service for VPA operator
* Remove vestigial webhook server
* Add operator E2E tests, including a test of centralized TLS profiles
* Cleanup: remove trailing whitespace
* Fixes for linter


maxcao13:
/lgtm

Thanks! I'm gonna follow up this PR because I believe we have a problem with the kubebuilder rbac markers, and our config/bundle rbac are not up to date.
Ref: https://github.com/kubernetes-sigs/controller-tools/issues/551
Which I've no clue why this is still an issue after 5 years.

joelsmith:
> I'm gonna follow up this PR because I believe we have a problem with the kubebuilder rbac markers, and our config/bundle rbac are not up to date.
> Ref: [kubernetes-sigs/controller-tools#551](https://github.com/kubernetes-sigs/controller-tools/issues/551)
> Which I've no clue why this is still an issue after 5 years.

Wow, mystery solved! I have always wondered why our rbac stuff didn't auto-update, but never spent much time looking into it. Thanks for getting to the bottom of it

PR #231, opened 2026-06-08
AUTOSCALE-283: annotation guard to prevent re-creation of default vpa controller

fishereskew:
Addresses [AUTOSCALE-283](https://redhat.atlassian.net/jira/software/c/projects/AUTOSCALE/boards/10228?selectedIssue=[AUTOSCALE-283](https://redhat.atlassian.net/browse/AUTOSCALE-283))

- Operator adds namespace annotation (vertical-pod-autoscaler-controller/created-default-controller) on startup, preventing duplicate creation of a default vpa controller on operator restart or controller deletion.
- Added RBAC permissions for namespace get/list/patch/watch.
- Added envTest integration tests to prevent future regressions.
- Added dependencies for envTest.


bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
```diff
@@ -41,7 +41,8 @@ metadata:
           "apiVersion": "autoscaling.openshift.io/v1",
           "kind": "VerticalPodAutoscalerController",
           "metadata": {
-            "name": "default"
+            "name": "default",
+            "namespace": "openshift-vertical-pod-autoscaler"
```
fishereskew, line 45: Only added this so that the `make create_vpa_controller_cr` command would correctly deploy the controller to the openshift-vertical-pod-autoscaler namespace.

joelsmith: :+1:
I'm surprised that we haven't run into issues with that makefile target before. Nice job figuring out the fix

joelsmith:
All of this looks great to me, thanks!
/approve

internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
```diff
@@ -57,6 +57,8 @@ import (
 const (
 	// ControllerName The hard-coded name of the VPA controller
 	ControllerName = "vertical-pod-autoscaler-controller"
+	// Operator namespace annotation for default controller creation
+	DefaultVPAControllerAnnotation = "vertical-pod-autoscaler-controller/created-default-controller"
```
maxcao13: Can we change the annotation to include the openshift autoscaling api group, since it's going to be part of a new API?

probably something like this?
```suggestion
	DefaultVPAControllerAnnotation = "autoscaling.openshift.io/default-vpa-controller-created"
```

Makefile
```diff
@@ -141,7 +141,7 @@ yamllint: ## Run yamllint against manifests.
 # TODO(macao): Future task to migrate to using envtest https://sdk.operatorframework.io/docs/building-operators/golang/testing/
```
maxcao13: Can you remove this TODO? Seems like you've integrated it well now :-)

maxcao13 (commented):
looks pretty good, just a few comments :-)

maxcao13:
/lgtm

thanks! a year in the making :smile:

fishereskew:
@maxcao13 what needs to be done for the verified label?

maxcao13:
If you think the envtests are sufficiently serving as regression tests, then you can do `/verified by unit` or `/verified by <whatever>`. If you think you need a QE person to look it over, you would ping them and ask them to verify the PR for you, however they choose to do so. 

But basically we just need the label by doing `/verified by <...>` and that `/verified by` can be anything you want.

e.g.

/verified by unit,@maxcao13

PR #232, opened 2026-06-10
OCPBUGS-87824: only use central TLS profile when the cluster says to

joelsmith:
Conditionally honor the centralized TLS profile based on APIServer.Spec.TLSAdherence:

* When `ShouldHonorClusterTLSProfile()` returns `false` as in the default case (`NoOpinion`) and `LegacyAdheringComponentsOnly`, the operator sets a nil TLSProfileSpec in the VPA controller config struct, so the admission plugin gets no `--min-tls-version` or `--tls-ciphers` args
* When `ShouldHonorClusterTLSProfile()` returns `true` as in the case of the TLS adherence policy `StrictAllComponents`, the operator passes on the cluster's centralized TLS profile (as added in #228)

The operator's struct Config.TLSProfileSpec is changed from value type to pointer, with nil meaning "use defaults, don't set TLS args."

This change breaks the E2Es added in #228, since all three existing tests assume the operator always honors the cluster TLS profile. In cases where there is no TLSAdherence policy or where we're told to ignore the profiles. Tests have been restructured into table-driven tests of 2 main flavors:

1. Sets TLSAdherence to StrictAllComponents, and basically runs the existing tests, but they're all skipped TLSAdherence feature gate is not enabled.
2. Sets TLSAdherence to NoOpinion (which is the default). These tests always run since they have the same behavior with or without the feature gate. They verify that the operator does not set TLS args, even when a custom TLS profile is configured (2 tests).


joelsmith:
Infra flake
/test e2e-aws-operator-components

maxcao13:
/lgtm

makes sense to me, thanks!

PR #235, opened 2026-06-17
Add APIServer config CRD for CI tests

joelsmith:
To fix failures of the flavor
```
 KUBEBUILDER_ASSETS="/go/src/github.com/openshift/vertical-pod-autoscaler-operator/bin/k8s/1.30.3-linux-amd64" go test $(go list ./... | grep -v /e2e) -coverprofile cover.out
	github.com/openshift/vertical-pod-autoscaler-operator/api/v1		coverage: 0.0% of statements
	github.com/openshift/vertical-pod-autoscaler-operator/cmd		coverage: 0.0% of statements
	github.com/openshift/vertical-pod-autoscaler-operator/cmd/testutil/json2yaml		coverage: 0.0% of statements
	github.com/openshift/vertical-pod-autoscaler-operator/cmd/testutil/yaml2json		coverage: 0.0% of statements
Running Suite: Controller Suite - /go/src/github.com/openshift/vertical-pod-autoscaler-operator/internal/controller/verticalpodautoscaler
=========================================================================================================================================
Random Seed: 1781674692
Will run 3 of 3 specs
2026-06-17T05:38:14Z	ERROR	controller-runtime.source.Kind	if kind is a CRD, it should be installed before calling Start	{"kind": "APIServer.config.openshift.io", "error": "no matches for kind \"APIServer\" in version \"config.openshift.io/v1\""}
sigs.k8s.io/controller-runtime/pkg/internal/source.(*Kind[...]).Start.func1.1
	/go/src/github.com/openshift/vertical-pod-autoscaler-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/source/kind.go:75
k8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext.func1
	/go/src/github.com/openshift/vertical-pod-autoscaler-operator/vendor/k8s.io/apimachinery/pkg/util/wait/loop.go:53
k8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext
	/go/src/github.com/openshift/vertical-pod-autoscaler-operator/vendor/k8s.io/apimachinery/pkg/util/wait/loop.go:54
k8s.io/apimachinery/pkg/util/wait.PollUntilContextCancel
	/go/src/github.com/openshift/vertical-pod-autoscaler-operator/vendor/k8s.io/apimachinery/pkg/util/wait/poll.go:33
sigs.k8s.io/controller-runtime/pkg/internal/source.(*Kind[...]).Start.func1
	/go/src/github.com/openshift/vertical-pod-autoscaler-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/source/kind.go:68
I0617 05:38:14.909690   37735 verticalpodautoscaler_controller.go:490] No VerticalPodAutoscalerController exists. Creating instance 'openshift-vertical-pod-autoscaler/default'
••
------------------------------
• [FAILED] [5.005 seconds] 
```

Also fix flakes in `VerticalPodAutoscalerController Controller VPA Controller creation upon startup [It] should add annotation to namespace and not create default controller when controller exists on startup ` by wrapping it in an eventually

### DIFF
--- a/internal/controller/verticalpodautoscaler/suite_test.go
+++ b/internal/controller/verticalpodautoscaler/suite_test.go
@@ -89,7 +89,10 @@ var _ = AfterSuite(func() {
 
 func StartEnvTest() (*rest.Config, *envtest.Environment, error) {
 	testEnv := &envtest.Environment{
-		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "..", "config", "crd", "bases")},
+		CRDDirectoryPaths: []string{
+			filepath.Join("..", "..", "..", "config", "crd", "bases"),
+			filepath.Join("..", "..", "..", "test", "testdata", "crd"),
+		},
 		ErrorIfCRDPathMissing: true,
 	}
 
--- a/.yamllint
+++ b/.yamllint
@@ -19,5 +19,6 @@ ignore:
 - config/vpa/vpa-crd.yaml
 - config/vpa/vpa-rbac.yaml
 - config/vpa/vpacheckpoint-crd.yaml
+- test/testdata/crd/
 
 # we need to ignore vpa yamls because manifest-diff-upstream copies it directly from the upstream repo and expects no git diff
--- a/Makefile
+++ b/Makefile
@@ -138,6 +138,10 @@ manifest-diff: build-testutil ## Compare permissions and CRDs from upstream mani
 yamllint: ## Run yamllint against manifests.
 	hack/yaml-lint.sh
 
+.PHONY: fetch-test-crds
+fetch-test-crds: yq ## Download OpenShift CRDs needed for envtest.
+	hack/fetch-test-crds.sh
+
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path --use-deprecated-gcs=false)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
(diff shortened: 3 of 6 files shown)


internal/controller/verticalpodautoscaler/suite_test.go
```diff
@@ -101,3 +104,48 @@ func StartEnvTest() (*rest.Config, *envtest.Environment, error) {
 
 	return cfg, testEnv, nil
 }
+
+// apiServerCRD returns a minimal APIServer CRD for use in envtest.
+// The real CRD is provided by the OpenShift config-operator at runtime;
+// this stub is just enough to let the controller's informer start.
+func apiServerCRD() *apiextensionsv1.CustomResourceDefinition {
```
maxcao13: Would it be too much to store required crds in a testdata folder, and use `CRDDirectoryPaths` instead of hardcoding a minimal APIServer CRD?

I did some claude digging and notice other openshift repos do similar things, e.g.
https://github.com/openshift/operator-framework-olm/blob/a78cfd39c259a22c104831c6ddf2572801b0f19d/staging/operator-lifecycle-manager/scripts/generate_openshift_crds.sh#L28
^ olm folk manage their own albatross envtest crd generation script

lvm-operator does something similar but don't maintain a script, so I'm assuming these crds are stale: https://github.com/openshift/lvm-operator/tree/main/test/e2e/testdata

If we don't feel like we will be depending on other crds in the envtests in the future or if you think it's not worth a rewrite, then I'm happy to just approve as is.

joelsmith: Okay, I have switched to this approach. My concern with doing it this way was ending up with stale CRDs (vs what I had which was a stub which wouldn't have any field to get out of date). But I agree that having the actual CRD is a better approach. So I added a hack script to be run during the chore cycle to get the latest CRD(s).

maxcao13:
/lgtm

Thanks!

PR #236, opened 2026-06-17
AUTOSCALE-67: VPA: Reduce potentially unsafe/uneeded privileges deployed by vpa operator

alaxkar:
As part of [AUTOSCALE-67](https://redhat.atlassian.net/browse/AUTOSCALE-67); we are Investigating the `vpa-rbac.yaml` to enforce the least privilege principle for the VPA.

1. This PR especially fixes below RBACS:
ClusterRole: `system:vpa-actor` 
ClusterRole: `system:vpa-target-reader` by removing wildcard entry
ClusterRole: `system:vpa-admission-controller` by specifically providing the webhook resourceName
ClusterRole: `system:vpa-checkpoint-actor` Removed (but added back) namespace `list` permission to satisfy failing tests.
ClusterRole: `system:vpa-updater-in-place` removed (but added back) `pods` patch resource to satisfy failing tests.

2. Since the changes have been done at `vpa-rbac.yaml` file, which is in sync with the upstream RBAC file, we need to modify the `hack/manifest-diff-upstream.sh` script in such a way that if there is any addition/removal of RBAC in the upstream, it should also reflect in out `vpa-rbac.yaml` file. (i.e; we can't directly remove the RBAC sync from the manifest diff script).

### DIFF
--- a/bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
+++ b/bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
@@ -56,7 +56,7 @@ metadata:
     categories: OpenShift Optional
     certifiedLevel: Primed
     containerImage: quay.io/openshift/origin-vertical-pod-autoscaler-operator:4.22.0
-    createdAt: "2026-06-18T07:51:54Z"
+    createdAt: "2026-06-19T14:27:08Z"
     description: An operator to run the OpenShift Vertical Pod Autoscaler. Vertical
       Pod Autoscaler (VPA) can be configured to monitor a workload's resource utilization,
       and then adjust its CPU and memory limits by updating the pod (future) or restarting
@@ -504,6 +504,7 @@ spec:
           - namespaces
           verbs:
           - get
+          - list
         - apiGroups:
           - autoscaling.k8s.io
           resources:
@@ -647,6 +648,7 @@ spec:
         - apiGroups:
           - ""
           resources:
+          - pods
           - pods/resize
           verbs:
           - patch
--- a/bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
+++ b/bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
@@ -56,7 +56,7 @@ metadata:
     categories: OpenShift Optional
     certifiedLevel: Primed
     containerImage: quay.io/openshift/origin-vertical-pod-autoscaler-operator:4.22.0
-    createdAt: "2026-06-20T06:10:09Z"
+    createdAt: "2026-06-24T14:04:09Z"
     description: An operator to run the OpenShift Vertical Pod Autoscaler. Vertical
       Pod Autoscaler (VPA) can be configured to monitor a workload's resource utilization,
       and then adjust its CPU and memory limits by updating the pod (future) or restarting
@@ -395,6 +395,13 @@ spec:
           - list
           - watch
         - apiGroups:
+          - '*'
+          resources:
+          - '*/scale'
+          verbs:
+          - get
+          - watch
+        - apiGroups:
           - ""
           resources:
           - replicationcontrollers
@@ -460,6 +467,8 @@ spec:
           - get
           - list
           - watch
+          - patch
+          - update
         - apiGroups:
           - poc.autoscaling.k8s.io
           resources:
@@ -512,6 +521,13 @@ spec:
           verbs:
           - get
           - patch
+        - apiGroups:
+          - '*'
+          resources:
+          - '*/scale'
+          verbs:
+          - get
+          - watch
         - apiGroups:
           - ""
           resources:
@@ -571,6 +587,8 @@ spec:
           - get
           - list
           - watch
+          - patch
+          - update
         - apiGroups:
           - poc.autoscaling.k8s.io
           resources:
@@ -607,6 +625,13 @@ spec:
           verbs:
           - get
           - list
+          - watch
+        - apiGroups:
+          - '*'
+          resources:
+          - '*/scale'
+          verbs:
+          - get
           - watch
         - apiGroups:
           - ""
--- a/bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
+++ b/bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
@@ -56,7 +56,7 @@ metadata:
     categories: OpenShift Optional
     certifiedLevel: Primed
     containerImage: quay.io/openshift/origin-vertical-pod-autoscaler-operator:4.22.0
-    createdAt: "2026-06-17T14:00:38Z"
+    createdAt: "2026-06-18T07:51:54Z"
     description: An operator to run the OpenShift Vertical Pod Autoscaler. Vertical
       Pod Autoscaler (VPA) can be configured to monitor a workload's resource utilization,
       and then adjust its CPU and memory limits by updating the pod (future) or restarting
@@ -504,7 +504,6 @@ spec:
           - namespaces
           verbs:
           - get
-          - list
         - apiGroups:
           - autoscaling.k8s.io
           resources:
--- a/bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
+++ b/bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
@@ -56,7 +56,7 @@ metadata:
     categories: OpenShift Optional
     certifiedLevel: Primed
     containerImage: quay.io/openshift/origin-vertical-pod-autoscaler-operator:4.22.0
-    createdAt: "2026-06-19T14:27:08Z"
+    createdAt: "2026-06-20T06:10:09Z"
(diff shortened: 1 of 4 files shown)


alaxkar:
All tests are now passing except test: `ci/prow/manifest-diff` and  is failing with error:
```
hack/manifest-diff-upstream.sh failed. Permissions not equivalent in config/vpa/vpa-rbac.yaml and /tmp/manifest-diff.kaylThaPUM/upstream/vpa-rbac.yaml
If changes are made to /tmp/manifest-diff.kaylThaPUM/upstream/vpa-rbac.yaml, equivalent changes should be made to config/vpa/vpa-rbac.yaml
If OpenShift-specific changes are made to config/vpa/vpa-rbac.yaml, those changes should be represented in hack/filter-upstream-rbac.jq so that a transformed upstream rbac will match the downstream changed version.
make: *** [Makefile:134: manifest-diff] Error 1
{"component":"entrypoint","error":"wrapped process failed: exit status 2","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:84","func":"sigs.k8s.io/prow/pkg/entrypoint.Options.internalRun","level":"error","msg":"Error executing test process","severity":"error","time":"2026-06-19T18:36:46Z"} 
```
_This seems reasonable to me, as it is validating `config/vpa/vpa-rbac.yaml` against the upstream RBAC definitions. 
The objective of this issue is to address the additional privileged RBAC permissions that were previously synchronized from upstream, so the current behavior aligns with that goal. 
With this fix in place, we also have an opportunity to revisit and enhance our test coverage._

**Update**: My bad, we should still design the `hack/manifest-diff-upstream.sh` script in such a way that all changes to the upstream RBACs should be flagged/synchronised with the current `vpa-rbac.yaml` file. In the next commit I will make the changes accordingly.

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
```diff
@@ -386,13 +394,6 @@ spec:
           - get
           - list
           - watch
-        - apiGroups:
-          - '*'
-          resources:
-          - '*/scale'
-          verbs:
-          - get
-          - watch
```
maxcao13, line 395: I'm not sure we can fully get rid of this on any of the components. If someone is using VPA with a custom resource that implements the /scale subresource, VPA wouldn't be able to scale upon it. It was introduced a long time ago, but I think it's still relevant: https://github.com/kubernetes/autoscaler/pull/3139

And see the faq here: https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/docs/faq.md#how-can-i-apply-vpa-to-my-custom-resource

Side note, I'm wondering if it makes sense to introduce new e2e tests to the upstream that test VPA scaling a custom resource. I don't think that exists today, and if we removed this rbac I think it would fail. Out of scope here though.

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
```diff
@@ -356,10 +356,18 @@ spec:
...
+        - apiGroups:
+          - admissionregistration.k8s.io
+          resourceNames:
+          - vpa-webhook-config
+          resources:
+          - mutatingwebhookconfigurations
+          verbs:
+          - delete
           - patch
+          - update
```
maxcao13, line 370: I like this change! I think this would be good to be proposed to the upstream to tighten the example rbac. Then we wouldn't have to maintain the added hacks in `filter-upstream-rbac.jq`.

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
maxcao13: Actually I see a bunch of error logs regardless:
```
E0623 00:25:11.338848 1 api.go:175] "Failed to get parent controller for pod" err="unhandled targetRef operator.openshift.io/v1 / DNS / default, last error dnses.operator.openshift.io \"default\" is forbidden: User \"system:serviceaccount:openshift-vertical-pod-autoscaler:vpa-updater\" cannot get resource \"dnses/scale\" in API group \"operator.openshift.io\" in the namespace \"openshift-dns\"" pod="openshift-dns/dns-default-hcz6l"
```
It's because the VPA is constantly scanning all pods in the cluster and trying to read their parent controllers scale subresource to check whether VPA can act upon it.

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
```diff
@@ -579,8 +571,6 @@ spec:
           - get
           - list
           - watch
-          - patch
-          - update
```
maxcao13, line 583: Did you investigate and find out we don't need this anymore? I think this was introduced last year because EventRecorder patches and updates Event objects: https://github.com/kubernetes/autoscaler/pull/8091

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
```diff
@@ -459,8 +460,6 @@ spec:
...
```
maxcao13, line 463: You're right that recommender probably doesn't need this. Actually looking at it, I think some of the rbac that's shared between updater and recommender can be tightened as well. I don't think recommender does anything with Events.

It might be worthwhile to investigate what is needed to be shared between updater and recommender and what isn't. I think the `system:vpa-actor` role is bound to both updater and recommender, when some of the perms do not overlap. This _might_ be worthwhile to contribute upstream, so we don't have to maintain hacks in the operator to do so.

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
joelsmith: Yes, Max is correct We must keep this permission.

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
```diff
@@ -513,13 +512,6 @@ spec:
           verbs:
           - get
           - patch
-        - apiGroups:
-          - '*'
-          resources:
-          - '*/scale'
-          verbs:
-          - get
-          - watch
```
joelsmith, line 522: This one needs to stay

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
```diff
@@ -618,13 +608,6 @@ spec:
           - get
           - list
           - watch
-        - apiGroups:
-          - '*'
-          resources:
-          - '*/scale'
-          verbs:
-          - get
-          - watch
```
joelsmith, line 627: This one needs to stay

config/vpa/vpa-rbac.yaml
```diff
@@ -209,13 +207,6 @@ kind: ClusterRole
 metadata:
   name: system:vpa-target-reader
 rules:
-  - apiGroups:
-    - '*'
-    resources:
-    - '*/scale'
-    verbs:
-    - get
-    - watch
```
joelsmith, line 218: this one needs to stay

hack/yamls2list.sed
```diff
@@ -1,7 +1,11 @@
 # Transform concatenated yaml files of k8s objects into a k8s List
 # Add kind, metadata and items at the top
-1 i \
----\nkind: List\nmetadata: {}\napiVersion: v1\nitems:
+1 i\
+---\
+kind: List\
+metadata: {}\
+apiVersion: v1\
+items:
```
joelsmith, line 8: Nice readability improvement!

joelsmith (commented):
I'm glad to see that the RBAC was already in pretty good shape! I think once the scale subresource rules are added back and Max's questions are answered, we can merge this.

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
alaxkar: You are right, actually I checked the code and couldn't find the `patch` & `update`  verbs directly used by the serviceAccounts that's why I removed it. But as per the issue you have mentioned here, it needs to stay here. I am reverting it back.

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
alaxkar: Got it, reverting back this to the original one. Thank you!!

bundle/manifests/vertical-pod-autoscaler.clusterserviceversion.yaml
alaxkar: Added it back due to this issue: [kubernetes/autoscaler#8091](https://github.com/kubernetes/autoscaler/pull/8091) and reverted back to your next comment.
Both changes are inherited from a single change in `vpa-rbac.yaml` file so both will be updated together.

`You are right, actually I checked the code and couldn't find the patch & update verbs directly used by the serviceAccounts that's why I removed it. But as per the issue you have mentioned here, it needs to stay here. I am reverting it back.`

joelsmith:
/lgtm
/approve

Thanks!

PR #242, opened 2026-06-24
[4.20] OCPBUGS-92046: Fix in-place RBAC & CRD manifests & disable in-place E2Es

joelsmith:
It looks like we did the 4.20 operator rebase before the operand, so the manifest check passed at the time of the rebase but is failing now. The changes won't have any effect except to enable the feature-gated feature to do in-place updates in 4.20's VPA. Also, we should be skipping the InPlaceOrRecreate tests since their detection method is fragile and doesn't work for OpenShift.

Without these fixes the manifest checks and the E2Es fail.

### DIFF
--- a/bundle/manifests/autoscaling.k8s.io_verticalpodautoscalercheckpoints.yaml
+++ b/bundle/manifests/autoscaling.k8s.io_verticalpodautoscalercheckpoints.yaml
@@ -219,7 +219,7 @@ spec:
                 type: string
             type: object
         type: object
-    served: true
+    served: false
     storage: false
 status:
   acceptedNames:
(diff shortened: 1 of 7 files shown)


alaxkar:
/lgtm
Thank you! The changes LGTM, at `config/vpa`: disabled serving v1beta2 versions making it deprecated in both vpa and vpacheckpoint files; and to match with the upstream given `InPlaceOrRecreate` functionality which allows the updater to in-place pod resize. That leads to some additional rbac changes for the events creation and additional new ClusterRole/binding  for the vpa-updater

joelsmith:
This is not a bug but rather a CI only fix
/override jira/valid-bug

PR #250, opened 2026-07-23
AUTOSCALE-907: Use operator ns if WATCH_NAMESPACE isn't provided

joelsmith:
Since there is no watch namespace in OLMv1, we should fall back to the operator's namespace if no watch namespace is provided. But there are still users of the older OLM API whose VPA we still want to avoid breaking, so if the watch namespace is provided, continue to use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,8 @@ The operator manages the following custom resource:
   controls the configuration of the cluster's VPA 3 controller instances.
   The operator will only respond to the VerticalPodAutoscalerController resource named "default" in the
   managed namespace, i.e. the value of the `WATCH_NAMESPACE` environment
-  variable.  ([Example][VerticalPodAutoscalerController])
+  variable.  ([Example][VerticalPodAutoscalerController]). If the `WATCH_NAMESPACE` environment is not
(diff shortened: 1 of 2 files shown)


joelsmith:
Tested on a cluster to make sure that a missing annotation would result in a running pod with an empty var:
```
$ cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Pod
metadata:
  name: test-busybox
spec:
  containers:
  - name: busybox
    image: busybox:latest
    command: ["sleep", "infinity"]
    env:
    - name: WATCH_NAMESPACE
      valueFrom:
        fieldRef:
          fieldPath: metadata.annotations['olm.targetNamespaces']
    securityContext:
      allowPrivilegeEscalation: false
      runAsNonRoot: true
      runAsUser: 1000
      seccompProfile:
        type: RuntimeDefault
      capabilities:
        drop:
        - ALL
EOF

$ oc exec -ti test-busybox -- sh -c "strings /proc/1/environ" | grep WATCH_NAMESPACE=
WATCH_NAMESPACE=
```

maxcao13:
/lgtm

Nice, thanks for the cleanup!

PR #253, opened 2026-09-17
AUTOSCALE-66: Track CA cert hash to restart admission controller on rotation

alaxkar:
## Summary

Track CA cert hash to automatically restart VPA admission controller on service-ca rotation. Hash stored as pod template annotation triggers rolling restart when changed.

## Changes

- Add `syncCACertHash()` to compute SHA-256 hash during reconciliation
- Add `UpdateAdmissionControllerAnnotations()` for admission controller pod template
- Update `UpdateAutoscaler()` to detect hash changes
- Only affects admission controller (recommender/updater unchanged)


internal/controller/verticalpodautoscaler/verticalpodautoscaler_controller.go
```diff
@@ -421,6 +431,40 @@ func (r *VerticalPodAutoscalerControllerReconciler) syncTLSProfile(ctx context.C
...
 }
 
+// syncCACertHash fetches the CA certificate ConfigMap and updates the hash in the config.
+// If the fetch or hash computation fails, the existing hash is retained and a warning is logged.
+func (r *VerticalPodAutoscalerControllerReconciler) syncCACertHash(ctx context.Context) {
+	cmNN := types.NamespacedName{
+		Name:      CACertConfigMapName,
+		Namespace: r.Config.Namespace,
+	}
+
```
maxcao13, line 441: I was thinking that instead of continuing to allow the admission controller itself to manage it's own mutatingwebhook, that instead the operator would annotate the service to allow service-ca operator to fully manage both the mutatingwebhook + CA certs? I crudely wrote this in the original ticket in a comment:
```
 * Then come back to this bug

 **** Allow the operator, not the VPA admission-controller, to create the admission webhook

 **** Use the service-ca operator to annotate the admission-controller webhook directly, instead of mounting the cert in a secret

 **** This will remove the need for the admission controller to patch, delete or even create webhooks at all
``` 

IMO that would be a cleaner solution if we were designing from scratch since we don't have to maintain our CA cert sync logic, or continue to use the hacky selfRegistration feature in upstream VPA admission controller, and we would eliminate the webhook create/update/delete rbac permission for the admission pod.

Is there a reason why this is not the route we use here? One thing I could think of is the potential migration issues when a user is upgrading their operator, but I haven't fully groked the consequences of if we were to go down that path, and if there would be an easy way to mitigate any consequences. The current implementation you have here would be the easiest way forward, but continues to build upon our hacks. Just wondering your thoughts here.

